PR #9, opened 2017-08-31
Adds the Enforcer, fixes Blueshield gun sprites and adds the Blue ticket

FlattestGuitar:
The Blue ticket lets the Blueshield pick their weapon as soon as they spawn in - the Enforcer, the adv. revolver or the Detective's revolver. We should probably buff the advanced revolver thing, but we can think about that later.

The enforcer fires rubber bullets by default, but you can also load it with any lethal .45 bullet. The regular bullets deal 5 brute and 80 stamina damage.

You can currently get more Enforcer bullets using an autolathe - 30k metal for 10 bullets. Way worse value compared to .38 speedloaders, especially considering that you can't put a used up speedloader back in there. RIP.

*also, you can load lethals easily enough instead of that rubber crap, but that might end in you ending up in perma*


SamHPurp:
Does it need a ticket? Just do what the null rod does?

Birdtalon:
Or the detective's revolver for that matter. Although that may just be an icon skin?

FlattestGuitar:
The null rod has strict subtypes which don't do all that much. Here it's massively different ways of operation and procs called on the three different guns. Impossible without a *load* of code that wouldn't be usable anywhere else.

The Detective's revolver uses the reskin system, which only handles the change of the way something looks, not any of its procs. The same system is now on the Enforcer, letting people choose what the gun looks like. Still, though, impossible to refactor into something that changes behavior without a load of work.

SamHPurp:
@Birdtalon Are you happy with all the changes?

Birdtalon:
👍 fine with me

PR #10, opened 2017-09-01
Makes 60fps default for new players

FlattestGuitar:
😱 😱 😱 😱 😱 😱 😱 😱 😱 😱


SamHPurp:
Are we certain this'll cause no issues?

FlattestGuitar:
There's actually already a config var that handles this instead and the checks are all there. This shouldn't cause issues, but we might as well just change the config entry to keep everything without merge conflicts in the future. Blep.

PR #12, opened 2017-09-01
Broken bone changes

FlattestGuitar:
Broken bones now properly work on all carbons - mainly monkeys

The ability to break a bone is now properly set on limbs and will follow the limb should it get transplanted


code/modules/surgery/bodyparts/bodyparts.dm
```diff
@@ -143,6 +155,9 @@
 		if(ALIEN_BODYPART,LARVA_BODYPART) //aliens take double burn
 			burn *= 2
 
+	if(prob(brute*2))
```
Birdtalon: I'd like to see something a bit more comprehensive than just a percentage chance ideally. Only high-damage weapons should break bones in my opinion. Maybe something like `if(brute >= 15 && prob(brute*2))` stabbing someone with a scalpel should not break bones.

code/modules/surgery/bodyparts/bodyparts.dm
```diff
@@ -104,10 +105,21 @@
...
+	if(!has_bones)
+		return 0
+	return 1
+
+/obj/item/bodypart/proc/break_bone()
+	if(!can_break_bone())
 		return
 	broken = TRUE
+	spawn(1)//because otherwise it pops before the punch message; we don't want that
+		owner.visible_message("<span class='userdanger'>You hear a cracking sound coming from [owner]'s [parse_zone(src)].</span>", "<span class='warning'>You feel something crack in your [parse_zone(src)]!</span>", "<span class='warning'>You hear an awful cracking sound.</span>")
```
Birdtalon: Maybe we can find some nice sound effects for this later?

SamHPurp:
As disgusting as it'd be, I also would really like a sound effect for the bone breaking. What do you guys think?

FlattestGuitar:
I experimented a bit with it earlier. The issue is that every single item in the game makes a sound when you hit people with it. That sound will be way louder than the bone break unless we want to make the bone break something you can hear from half across the station. I'm not sure how to implement it otherwise. The laws of physics are against us.

SamHPurp:
Make it play only to the recipient on a slight delay?

FlattestGuitar:
The previous version played with a 0.3 second delay and it sounded really weird. It was audible, though. I can put it in if you want me to, your call.

Birdtalon:
Make it on no delay I would say. As long as the sound plays it should be recognisable.

PR #17, opened 2017-09-07
Removes firing pins

FlattestGuitar:
This one hurt. Ouch.

Have fun with the review, everyone.


SamHPurp:
@FlattestGuitar you didn't map merge half of those dmms.

SamHPurp:
Also we need the replacement solution to the problem: Lock boxes.

PR #21, opened 2017-09-14
Cryopods!

FlattestGuitar:
- not mapped yet, because I'm a bad mapper and we still might get around to changing more
- not tested like AT ALL and that will have to be done

Not really different from the Paradise implementation, except it cuts down on quite a bit of weird behavior and shitcode. 

Cryostorage will warn people who are either important antags or heads of staff. 

Both borgs and regular crewmembers can use the pods. No idea why Paradise forces people to go to the AI sat if they're borgs. The bit for cyborgs is commented out.

I can *guarantee* you that most of this crap doesn't work, but feel free to take a look at it since it's already written.


code/controllers/subsystem/job.dm
```diff
@@ -86,6 +86,14 @@ SUBSYSTEM_DEF(job)
 	Debug("AR has failed, Player: [player], Rank: [rank]")
 	return 0
 
+/datum/controller/subsystem/job/proc/FreeRole(rank)
+	if(!rank)
+		return
+	Debug("Freeing role: [rank]")
```
SamHPurp: Will need the debug code gone: yaya, nono?

code/controllers/subsystem/job.dm
```diff
@@ -86,6 +86,14 @@ SUBSYSTEM_DEF(job)
...
+	var/datum/job/job = GetJob(rank)
+	if(!job)
+		return 0
```
SamHPurp: `FALSE`

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,464 @@
...
+
+//Main cryopod console.
+
+/obj/machinery/computer/cryopod
+	name = "cryogenic oversight console"
+	desc = "An interface between crew and the cryogenic storage oversight systems."
+	icon = 'icons/obj/Cryogenic2.dmi'
+	icon_state = "cellconsole"
+	circuit = /obj/item/circuitboard/cryopodcontrol
+	density = 0
```
SamHPurp: `FALSE`

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,464 @@
...
+	interact_offline = 1
```
SamHPurp: `TRUE`

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,464 @@
...
+	req_one_access = list(ACCESS_HEADS, ACCESS_ARMORY) //Heads of staff or the warden can go here to claim recover items from their department that people went were cryodormed with.
+	var/mode = null
+
+	//Used for logging people entering cryosleep and important items they are carrying.
+	var/list/frozen_crew = list()
+	var/list/frozen_items = list()
+
+	var/storage_type = "crewmembers"
+	var/storage_name = "Cryogenic Oversight Control"
+	var/allow_items = 1
```
SamHPurp: `TRUE`

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,464 @@
...
+		/obj/item/areaeditor/blueprints,
+		/obj/item/clothing/head/helmet/space,
+		/obj/item/clothing/suit/space,
+		/obj/item/clothing/suit/armor,
+		/obj/item/defibrillator/compact,
+		/obj/item/reagent_containers/hypospray/CMO,
+		/obj/item/clothing/accessory/medal/gold/captain,
+		/obj/item/clothing/gloves/krav_maga,
+		/obj/item/storage/internal,
+		/obj/item/nullrod
```
SamHPurp: With some paths changing on TG, this may cause issues potentially.

Also, if there are new theft objectives... like pretty sure the Captain's got a jetpack which is a theft item now? Can you double check these.

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,464 @@
...
+		last_no_computer_message = world.time
+
+	return control_computer != null
+
+/obj/machinery/cryopod/close_machine(mob/user)
+	if((isnull(user) || istype(user)) && state_open && !panel_open)
+		..(user)
+		var/mob/living/mob_occupant = occupant
+		if(mob_occupant && mob_occupant.stat != DEAD)
+			to_chat(occupant, "<span class='notice'><b>You feel cool air surround you. You go numb as your senses turn inward.</b></span>")
```
SamHPurp: use `boldnotice` so you can drop the `<b>` tags

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,464 @@
...
+		if(mob_occupant.client)//if they're logged in
+			despawn_world_time = world.time + (time_till_despawn * 0.1)
+		else
+			despawn_world_time = world.time + time_till_despawn
+	icon_state = "cryopod"
+
+/obj/machinery/cryopod/open_machine()
+	..()
+	icon_state = "cryopod-open"
+	//TODO?
```
SamHPurp: 🤔

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,464 @@
...
+			if(!istype(W, T))
+				if(control_computer && control_computer.allow_items)
+					control_computer.frozen_items += W
+					W.loc = null
+				else
+					W.forceMove(loc)
+			else
+				qdel(W)
+	if(istype(SSticker.mode, /datum/game_mode/cult))//thank
+		if("sacrifice" in SSticker.mode.cult_objectives)
```
SamHPurp: Will this trigger for all cult forms? Clock?

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,464 @@
...
+		to_chat(user, "<span class='notice'>Dead people can not be put into cryo.</span>")
+		return
+
+	if(target.client && user != target)
+		to_chat(user, "<span class='notice'>You can't put conscious people into [src]!</span>")
+	else if(target.client)
+		if(alert(target,"Would you like to enter cryosleep?",,"Yes","No") == "No")
+			return
+
+	var/generic_plsnoleave_message = " Please adminhelp before leaving the round, unless there are no administrators online."
```
SamHPurp: `unless` -> `even`

We'll be outputting ahelps to discord... or the mins could be invisimin

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,464 @@
...
+	name = "[name] ([occupant.name])"
+	log_admin("<span class='notice'>[key_name(target)] entered a stasis pod.</span>")
+	message_admins("[key_name_admin(target)] entered a stasis pod. (<A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[x];Y=[y];Z=[z]'>JMP</a>)")
+	add_fingerprint(target)
+
+//Attacks/effects.
+/obj/machinery/cryopod/blob_act()
+	return //Sorta gamey, but we don't really want these to be destroyed.
+
+// /obj/machinery/computer/cryopod/robot
```
SamHPurp: Kinda need it for cyborgs.

code/controllers/subsystem/job.dm
FlattestGuitar: Not that sort of a debug. This is just for logging. It stays. >:C

code/game/machinery/cryopod.dm
FlattestGuitar: 😇

code/game/machinery/cryopod.dm
FlattestGuitar: Clock cult doesn't get any objectives. Just grow and "summon".

code/game/machinery/cryopod.dm
FlattestGuitar: The point is to let Cyborgs use regular cryo instead of the AI sat. It's more easily accessible.

I can still add the robot storage if you want me to, but I think it's a wasted effort...

Birdtalon: 👌

SamHPurp: Have you tested the regular ones working properly, and it not storing anything it shouldn't?

If so - lets remove all this commented out code entirely.

code/game/machinery/cryopod.dm
AndrewMontagne: Might be worth finding a better way of enumerating these in the future.

code/controllers/subsystem/job.dm
```diff
@@ -86,6 +86,14 @@ SUBSYSTEM_DEF(job)
...
 	return 0
 
+/datum/controller/subsystem/job/proc/FreeRole(rank)
+	if(!rank)
+		return
+	Debug("Freeing role: [rank]")
+	var/datum/job/job = GetJob(rank)
+	if(!job)
+		return FALSE
+	job.current_positions--
```
AndrewMontagne: Do we need a lower bounds check on this variable?

FlattestGuitar: Well, logic would imply we do not, so I'll include one.

Birdtalon: Unless some mess up happens elsewhere in the code when it comes to assigning `current_positions` there should never be a case where this underflows. - Famous last words.

FlattestGuitar:
I'd love if someone was willing to sit down with me and just test use cases for this thing, because it'd be really surprising if I didn't miss anything here.

SamHPurp:
Yeah, that'd be a good idea @FlattestGuitar

FlattestGuitar:
We already tested stuff yesterday. All the issues we found have been addressed.

AndrewMontagne:
LGTM

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,433 @@
...
+
+/obj/machinery/computer/cryopod/attack_ai()
+	attack_hand()
+
+/obj/machinery/computer/cryopod/attack_hand(mob/user = usr)
+	if(stat & (NOPOWER|BROKEN))
+		return
+
+	user.set_machine(src)
+	add_fingerprint(usr)
```
SamHPurp: You're assigning `usr` to `user`, might as well remain consistent using the variable `user`?

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,433 @@
...
+		if(!allowed(user))
+			to_chat(user, "<span class='warning'>Access Denied.</span>")
+			return
+		if(!allow_items) return
+
+		if(frozen_items.len == 0)
+			to_chat(user, "<span class='notice'>There is nothing to recover from storage.</span>")
+			return
+
+		var/obj/item/I = input(usr, "Please choose which object to retrieve.","Object recovery",null) as null|anything in frozen_items
```
SamHPurp: again, using `usr` and `user` intermittently?

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,433 @@
...
+
+/obj/item/circuitboard/cryopodcontrol
+	name = "Circuit board (Cryogenic Oversight Console)"
+	build_path = "/obj/machinery/computer/cryopod"
+	origin_tech = "programming=1"
+
+//Cryopods themselves.
+/obj/machinery/cryopod
+	name = "cryogenic freezer"
+	desc = "A man-sized pod for entering suspended animation. Now for Cyborgs too!"
```
SamHPurp: Re-phrase the whole sentence: "Now for Cyborgs too!" - looks super tacked on.

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,433 @@
...
+		/obj/item/areaeditor/blueprints,
+		/obj/item/clothing/head/helmet/space,
+		/obj/item/clothing/suit/space,
+		/obj/item/clothing/suit/armor,
+		/obj/item/defibrillator/compact,
+		/obj/item/reagent_containers/hypospray/CMO,
+		/obj/item/clothing/accessory/medal/gold/captain,
+		/obj/item/clothing/gloves/krav_maga,
+		/obj/item/nullrod,
+		/obj/item/tank/jetpack,
```
SamHPurp: This includes a load of shit we don't want. Make this only the Captain's jetpack.

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,433 @@
...
+		/obj/item/documents
+	)
```
SamHPurp: You're missing the nuke core container and the supermatter bin objective items

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,433 @@
...
+	if(occupant)
+		to_chat(user, "<span class='boldnotice'>The cryo pod is already occupied!</span>")
+		return
+
+	if(target.stat == DEAD)
+		to_chat(user, "<span class='notice'>Dead people can not be put into cryo.</span>")
+		return
+
+	if(target.client && user != target)
+		to_chat(user, "<span class='danger'>You can't put conscious people into [src]!</span>")
```
SamHPurp: Says people - even if its a borg being pushed in

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,433 @@
...
+					reshape.Crop(7,4,26,31)
+					reshape.Crop(-5,-3,26,30)
+					GLOB.sac_image = reshape
+
+					for(var/datum/mind/H in SSticker.mode.cult)
+						if(H.current)
+							to_chat(H.current, "<span class='danger'>Nar'Sie</span> murmurs, <span class='cultlarge'>[occupant] is beyond your reach. Sacrifice [GLOB.sac_mind.current] instead...</span></span>")
+
+	//Update any existing objectives involving this mob.
+	for(var/datum/objective/O in GLOB.objectives)
```
SamHPurp: Silly edgecase - but if there are no other possible targets... it just keeps your original "assassinate Jeff" objective.

SamHPurp:
Attachable helmets (such as the Syndie commando helmet, which is attached to the hardsuit)... once you cryo - both items go into the cryogenic console...

When you pop out both items. You pick up the helmet - the helmet pops out of existence and appears on the armour.

SamHPurp:
Need a right click verb for Cryopods to eject the occupant.

code/game/machinery/cryopod.dm
AndrewMontagne: Would a variable on /obj/item denoting for it not to be cryo'ed, etc be more useful instead of holding a list of them someplace else? Maybe a bitmask, with one of the possibilities being DO_NOT_CRYO?

SamHPurp:
Items that pop out of the cryo console seem to have their layers fucked. I'm assuming they're inheriting or something the Cryo consoles layer.

code/game/machinery/cryopod.dm
FlattestGuitar: Jetpacks are very hard to obtain above what's already there at shift start. I don't see why we wouldn't want to catch the ones that end up in cryo.

SamHPurp: There are jetpack upgrades that go onto void suits - which then pop out of the suit and get kept separately, as well as having already upgraded the suit.... duplicating the item.

Thats what caused that concern.

code/game/machinery/cryopod.dm
```diff
@@ -0,0 +1,442 @@
...
+
+	for(var/obj/item/W in mob_occupant.GetAllContents())
+		if(W.loc.loc && (( W.loc.loc == loc ) || (W.loc.loc == control_computer)))
+			continue//means we already moved whatever this thing was in
+			//I'm a professional, okay
+		for(var/T in preserve_items)
+			if(istype(W, T))
+				if(control_computer && control_computer.allow_items)
+					control_computer.frozen_items += W
+					mob_occupant.transferItemToLoc(W, control_computer, TRUE)
```
FlattestGuitar: Now uses the proper unequip procs. Will hopefully make it not break layers, though I couldn't really replicate it myself. :v

code/game/gamemodes/objective.dm
```diff
@@ -38,6 +41,8 @@
 			possible_targets += possible_target
 	if(possible_targets.len > 0)
 		target = pick(possible_targets)
+	else
+		target = null//we'd rather have no target than an invalid one
```
FlattestGuitar: This proc was never used to assign a new target outside of what happens when the objective is created. This is safe.

code/game/machinery/cryopod.dm
FlattestGuitar: They'll be kept along with the hardsuit now, not separately.

code/game/machinery/cryopod.dm
```diff
@@ -329,14 +331,18 @@
 		announcer.announce("CRYOSTORAGE", mob_occupant.real_name, announce_rank, list())
 		visible_message("<span class='notice'>\The [src] hums and hisses as it moves [mob_occupant.real_name] into storage.</span>")
 
+
 	for(var/obj/item/W in mob_occupant.GetAllContents())
+		if(W.loc.loc && (( W.loc.loc == loc ) || (W.loc.loc == control_computer)))
+			continue//means we already moved whatever this thing was in
+			//I'm a professional, okay
```
SamHPurp: 🤔

SamHPurp:
Initially the changes look good. Andrew and I will do a bit more live-testing this evening.

SamHPurp:
Cryoing doesn't free the job slot

PR #30, opened 2017-09-18
Adds splints to brute medkits and medical vendors

FlattestGuitar:
Title.

They're not actually in the game currently.
Thanks, me.


SamHPurp:
We argued about this on Discord. I'm not sure where we got.

My stance is that there are too many.

FlattestGuitar:
The amount changed to 3 splints per stack. The standard Box medbay starts with 14 stacks all over the place.

Will be hard to gather all of them and I think it won't get out of hand. Hopefully?

PR #32, opened 2017-09-19
AI Late Joining

AndrewMontagne:
Does what it says on the tin. This PR depends on the wipe core PR being merged first.

:cl: AndrewMontagne
add: You can now join late as an AI if there's a deactivated core on the map.
/:cl:


AndrewMontagne:
[Add the ability to join late as an AI](https://trello.com/c/VwhUP8xy/63-add-the-ability-to-join-late-as-an-ai)

code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -330,6 +330,32 @@
...
+			continue
+		if(locate(/mob/living) in S.loc)
+			continue
+		GLOB.empty_playable_ai_cores += new /obj/structure/AIcore/deactivated(get_turf(S))
+
+	return 1
+
+// Before calling this, make sure an empty core exists, or this will no-op
+/mob/living/silicon/ai/proc/moveToEmptyCore()
+	if(!GLOB.empty_playable_ai_cores.len)
```
FlattestGuitar: This should probably throw an alert at admins or an exception or something. Will be a bitch to debug if we forget somewhere down the line.

code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -330,6 +330,32 @@
 /mob/living/silicon/ai/cancel_camera()
 	view_core()
 
+GLOBAL_LIST_EMPTY(empty_playable_ai_cores)
+
+/proc/roundstart_spawn_empty_ai_if_needed()
+	for(var/obj/effect/landmark/start/S in GLOB.landmarks_list)
+		if(S.name != "AI")
+			continue
+		if(locate(/mob/living) in S.loc)
```
FlattestGuitar: Should also check for pre-existing AI cores.

code/modules/mob/transform_procs.dm
```diff
@@ -346,6 +346,7 @@
 	. = new /mob/living/silicon/ai(loc_landmark, null, src)
 
 	qdel(src)
+	return .
```
FlattestGuitar: `.` refers to the return value. All functions return `.` unless specifically told to return something else. This is redundant.

FlattestGuitar:
When shit breaks, it will be good to have either an admin verb or just a simple proc we can call on an empty core that lets us make a new empty AI at a different spot.

code/modules/mob/living/silicon/ai/ai.dm
SamHPurp: isn't it already doing that?

SamHPurp:
What happens if there is an AI in a core.... I card him.... can someone then just late join into that core?... We probably want to stop that if so!

FlattestGuitar:
LGTM.

PR #36, opened 2017-09-19
Set default client FPS to 60.

AndrewMontagne:
Does exactly what it says on the tin.

:cl: AndrewMontagne
tweak: Set default client FPS to 60
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -100,7 +100,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 
 	var/list/ignoring = list()
 
-	var/clientfps = 0
+	var/clientfps = 60
 
 	var/parallax
 


AndrewMontagne:
[Set client FPS to 60](https://trello.com/c/jdSgDqI4/66-set-client-fps-to-60)

SamHPurp:
Are we certain this isn't going to be breaking things?

AndrewMontagne:
@SamHPurp As far as I am aware, the values of the object always get overwritten when they're loaded in from the persistent store. This is just the default of value upon construction, so if nothing gets loaded in, this is what the get. (See code/modules/client/preferences_savefile.dm:219)

FlattestGuitar:
Upon further inspection of the code, I have found this way of soliving things to be optimal. #10 was wrongly closed. </3

LGTM.

PR #41, opened 2017-09-20
Emote me up, Scotty!

FlattestGuitar:
- Disables CONTROL_FREAK on macros, so people can now use macros. .click is disabled
- Adds sounds to *scream, *cough and *sneeze
- Adds *yes and *no emotes
- Since the *screech emote for monkeys now makes a sound, xenos can no longer use it. Sorry, xenos.
- Adds a cooldown system for emotes, so you can't spam them


code/modules/mob/living/emote.dm
```diff
@@ -143,12 +143,12 @@
 /datum/emote/living/flip
 	key = "flip"
 	key_third_person = "flips"
+	message = "does a flip!"
```
SamHPurp: I could totally do without it making a message.

code/modules/mob/living/silicon/robot/emote.dm
```diff
@@ -70,3 +83,13 @@
 			playsound(loc, 'sound/machines/buzz-two.ogg', 50, 0)
 		else
 			to_chat(src, "<span class='warning'>You can only use this emote when you're out of charge.</span>")
+
+/datum/emote/sound/silicon/yes
+	key = "yes"
+	message = "blips affirmatively."
```
SamHPurp: blip is a bit more... discordant...?

Beep might be better.

FlattestGuitar: But it's a blip!

PR #47, opened 2017-09-22
Go-live Fixes

AndrewMontagne:
The pipeline warning was an extraneous warning, a proc to fix the warned issue is called right after. Fixed an oversight with the PDA slot changes. Made the duplicate radio messages go away, but no idea why /tg/ doesn't have this problem...?

:cl: AndrewMontagne
fix: Fixed PDA slot issue (#31)
fix: Fixed duplicate radio messages
/:cl:


SamHPurp:
OK so hopefully the changelog bot will work. Fingers crossed.

PR #51, opened 2017-09-23
Ports frost to LINDA.

AndrewMontagne:
Ported the frost that occurs when tiles are cold from Paradise.

image image
image

:cl: AndrewMontagne
add: Frost when tiles are below freezing.
/:cl:


AndrewMontagne:
[Port frost floor tile effect from Paradise.](https://trello.com/c/L5QXDf5B/61-port-frost-floor-tile-effect-from-paradise)

PR #52, opened 2017-09-24
Rubber ducky no make a squeak but a honk!

Birdtalon:
RUBBER DUCKY NO MAKE A HONK

🆑 Birdtalon
add: Adds rubber ducky squeak sound.
/🆑


FlattestGuitar:
*sad honk*

PR #72, opened 2017-09-27
Puts a leash on ash drakes

### LINKED ISSUE #65: Ash Drakes Fly Attack not dodgeable with current speed.
So considering we got lowered speed and we got megafauna I thought I would try out the easiest of them. Normally with enough speed the ashdrake is manageable. You can dodge all of his attacks with the current speed but one. Thats the flying sweep, you know where he flies up into the air and stomps you if you dont dodge in time. Usually you got a second to get away which is, provided you have normal TG movementspeed absolutly doable. As it stands now its pretty much impossible. Ash Drake used his sweep 3 times on me and couldnt dodge them at all because they are always targetting the center. I suggest to fix this minor issue the ashdrake should get a small delay on his sweep attack, say 1.5 seconds. That way you can dodge his sweep if you react in time which is (near) impossible right now. Other than that the megafauna is pretty much good to fight against at the speed we now have.

Birdtalon:
Increases the time taken for Ash Drake swoops to land to better meld with our slower move speed. They should still be a challenge without being so close to the wire.

Fixes #65

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/megafauna/dragon.dm
+++ b/code/modules/mob/living/simple_animal/hostile/megafauna/dragon.dm
@@ -107,7 +107,7 @@ Difficulty: Medium
 
 	if(prob(15 + anger_modifier) && !client)
 		if(health < maxHealth/2)
-			INVOKE_ASYNC(src, .proc/swoop_attack, TRUE, null, 50)
+			INVOKE_ASYNC(src, .proc/swoop_attack, TRUE, null, 75)
 		else
 			fire_rain()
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/hostile/megafauna/dragon.dm
```diff
@@ -154,11 +154,11 @@ Difficulty: Medium
...
 /mob/living/simple_animal/hostile/megafauna/dragon/proc/triple_swoop()
-	swoop_attack(swoop_duration = 30)
-	swoop_attack(swoop_duration = 30)
-	swoop_attack(swoop_duration = 30)
+	swoop_attack(swoop_duration = 50)
+	swoop_attack(swoop_duration = 50)
+	swoop_attack(swoop_duration = 50)
 
-/mob/living/simple_animal/hostile/megafauna/dragon/proc/swoop_attack(fire_rain, atom/movable/manual_target, swoop_duration = 40)
+/mob/living/simple_animal/hostile/megafauna/dragon/proc/swoop_attack(fire_rain, atom/movable/manual_target, swoop_duration = 80)
```
Birdtalon: Default duration changed to 80. Maybe 60 is better?

FlattestGuitar:
Works for me. We can adjust later if it's too easy.

SamHPurp:
[Make Lavaland playable for the slower speed](https://trello.com/c/Ac6Sh526/5-make-lavaland-playable-for-the-slower-speed)

PR #94, opened 2017-09-29
Bubblegum charge nerf

### LINKED ISSUE #84: Bubblegum charge not adjusted to reduced movement speed.
Same issue as with the ashdrake. You can react to the charge but the movement speed we got does not allow dodging in time. Its by far the most deadliest megafauna but should be doable without being a meth addict or having TGs movement. Havent had a chance to see his other attacks cause he decided to charge instantly with a speed of 300.000 kilometers per second. I suggest the same change like we gave the drake: Less speed on the charge to make it dodgeable without meth.

Birdtalon:
Nerfs the speed of bubblegum's charge in line with our slower movespeed. (60%)

Fixes #84


SamHPurp:
[Make Lavaland playable for the slower speed](https://trello.com/c/Ac6Sh526/5-make-lavaland-playable-for-the-slower-speed)

PR #100, opened 2017-09-29
Changes asay color to something perceptible by human sight.

FlattestGuitar:
BEFORE:

AFTER:

This color is also original and doesn't get confused with admin logs, of which there are A TON.


Xantholne:
Current Asay is literally the same color as **every, log, for everything**

PR #109, opened 2017-10-02
Fixes His Grace being aqquireable by surplus crate.

EldritchSigma:
Well first propsed change of mine. Found out that His Grace is surplusable and it really should be chaplin only. Fixed this quickly (and hopefully).

:cl: EldritchSigma
tweak: His Grace decided that greys with hot pink eyes are beneath his notice and decided that only holy chaplins can have him.
/ :cl:


EldritchSigma:
Ahhh I see. Fixed I think.

PR #116, opened 2017-10-09
"Fixes" titlescreen and makes R&D console more pleasant to use.

AffectedArc07:
**Title Screen Edits**
As you know, the current titlescreen is the default TG but with text removed, While I am sure this isnt going to be the permanent titlescreen, its a start for now:
image

**R&D Console Edits**
If you have played TG in the past, you will know that the RD console requires you to go into a seperate menu to view current R&D levels, not very convenient or practical. This has been fixed and the main menu displays them, similar to how para does it.
image

:cl: AffectedArc07
tweak: R&D Console now has current research levels on the main menu
imageadd: Title screen has add OracleStation added to it
/:cl:
In regards to why this should be in, I feel the title screen should be a good one for now even if it will be changed in the future, and the R&D console is just arguably nicer this way.


FlattestGuitar:
Please get rid of the title screen change. As I told you earlier, we're already working on that, so it's just needless merge conflicts. >:C

code/modules/research/rdconsole.dm
```diff
@@ -658,7 +658,6 @@ won't update every console in existence) but it's more of a hassle to do. Also,
 		if(1.0) //Main Menu
 			dat += "<div class='statusDisplay'>"
 			dat += "<h3>Main Menu:</h3><BR>"
-			dat += "<A href='?src=\ref[src];menu=1.1'>Current Research Levels</A><BR>"
```
SamHPurp: Why are you removing this.

AffectedArc07: That removes the link to the page which I have merged into the main menu, as for the title screen, how do I remove it?

PR #121, opened 2017-10-13
Removes old invisible vests. Rip.

FlattestGuitar:
As per title.

There are no longer two alternative vest styles to pick from, all the previously existing ones have been turned into the new version.

### DIFF
--- a/code/modules/clothing/suits/armor.dm
+++ b/code/modules/clothing/suits/armor.dm
@@ -14,21 +14,16 @@
 /obj/item/clothing/suit/armor/vest
 	name = "armor vest"
 	desc = "A slim Type I armored vest that provides decent protection against most types of damage."
-	icon_state = "armoralt"
-	item_state = "armoralt"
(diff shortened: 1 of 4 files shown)


code/modules/clothing/suits/armor.dm
```diff
@@ -14,21 +14,16 @@
...
-	item_state = "armoralt"
-	blood_overlay_type = "armor"
-	dog_fashion = /datum/dog_fashion/back
-
-/obj/item/clothing/suit/armor/vest/alt
-	desc = "A Type I armored vest that provides decent protection against most types of damage."
 	icon_state = "armor"
 	item_state = "armor"
+	blood_overlay_type = "armor"
+	dog_fashion = /datum/dog_fashion/back
```
SamHPurp: 😂

FlattestGuitar: Ian is my secsona.

PR #124, opened 2017-10-13
Configuration datum refactor 

FlattestGuitar:
Ports (#30763) from /tg/

All hail the config refactor.

It compiles and doesn't runtime at roundstart, so it's basically merge ready.


Anticept:
See https://github.com/tgstation/tgstation/pull/31252

Birdtalon:
To be honest with regards to @Anticept 's comment, the only thing I'd remove the protection from is the server name & hub flag. Everything else should really not be change-able while the server is running such as the SQL address, server URL etc.

FlattestGuitar:
We should at least remove the verbs that try to change the locked config vars. I'll sift through this when I have some time.

AndrewMontagne:
This is an awful lot to review manually. I would feel happier if this was testmerged and had a playtest run on it before merging into `master`.

FlattestGuitar:
All of these are something that might have to be changed mid-round if stuff breaks. I've had to do so myself on Paradise a couple of times, I doubt it will be different here.

All config bits can be changed, don't give admin to people you can't trust.

Birdtalon:
I would say it's not a case of trust but mistakes. But I can live with that.

SamHPurp:
Lets get this thoroughly tested so we can merge it.

SamHPurp:
[Port config datum refactor *Scream](https://trello.com/c/IHLjOh8Z/76-port-config-datum-refactor-scream)

SamHPurp:
:ohgod:

PR #125, opened 2017-10-15
Fix line ending

Allfd:
This breaks the server

:cl: alffd
fix: Removed a trailing . that broke byond 512
/:cl:


Birdtalon:
Yeah that probably shouldn't be there.

PR #135, opened 2017-10-16
Adds species printing on examine

FlattestGuitar:
Couldn't really find a clean way of implementing this madness without adding fourty different procs to species. There's `get_examine_text()` which will work if anyone needs to override the default one and there's a sexy var for all other applications.

Enjoy.


Birdtalon:
ethari

code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
```diff
@@ -19,6 +19,7 @@
 	disliked_food = GRAIN | DAIRY
 	liked_food = GROSS | MEAT
 	ass_pic = "unathi"
+	examine_text = "<font color='#3d0000'> a Unathi</font>"
```
SamHPurp: an

FlattestGuitar: The only codebase out there that has a sound file for Unathi is Paradise. It's "Yu-nathi" there, so that'd be an "a". >:C

SamHPurp:
I'M NOT HAPPY

PR #136, opened 2017-10-16
Removes player requirement for surplus crate

imsxz:
:cl: imsxz
tweak: Surplus crate can now be bought at any player count
/:cl:

I understand the logic behind locking the crate behind a player count, but part of the reason surplus crates are ordered is their potential to contain specific valuable items that may be locked behind other requirements such as needing to be a specific job.

This allows for the ability to acquire more valuable syndicate contraband that they otherwise wouldn't be able to get, by taking a risky gamble.

*It's also the only traitor item I feel comfortable buying*

### DIFF
--- a/code/modules/uplink/uplink_item.dm
+++ b/code/modules/uplink/uplink_item.dm
@@ -1384,7 +1384,6 @@ GLOBAL_LIST_EMPTY(uplink_items) // Global list so we only initialize this once.
 			but you never know. Contents are sorted to always be worth 50 TC."
 	item = /obj/structure/closet/crate
 	cost = 20
-	player_minimum = 20 //lowered, the times you get screwed over by this thing happens far too often
 	exclude_modes = list(/datum/game_mode/nuclear, /datum/game_mode/gang)
 	cant_discount = TRUE
 


Quantum-M:
I think the reason for such locking of the surplus crate depending the player count is that it can give you a potentially several powerful items. If there's a low population (most likely also low Security), then the antag would be too powerful and have a lower chance on being stopped if they go loud.

SamHPurp:
Might be easier to discuss over Discord, but, I am not in favour of this.

Particularly when you're low pop, your job as an antagonist is to make things interesting. You probably shouldn't be rolling the dice on that.

Birdtalon:
I am in favour. It should be the player's choice not dictated by the number of players IMO.

imsxz:
I can't speak for others, but I'm really shit at buying my own gear. When I'm left with my own decisions, it usually makes for a rather murderboney setup, or a rather super stealth setup. Surplus crate makes me need to think outside the box to put my gear to use.

Brasnak:
What about making a lesser surplus crate only containing 30-35 TC worth of syndicate gear that replaces the normal surplus crate during lowpop instead?

FlattestGuitar:
Honestly, I don't really know which way I want this to go myself.

The only sensible reason for which the surplus crate would be prohibited at low playercounts is the possibility of stupidly OP traitors popping up.

If we were to change the behavior of the crate to simply lower the TC amount somewhat if the playercount is below a certain level it'd be a fine solution to this.

imsxz:
If we're worried about people lucking out and getting OP traitor items in lowpop, I'd just like to mention that you can get His Grace from the "random item" button.

Additionally, you can get other items that require minimum amount of players from syndicate bundles. Syndicate bundles don't have a player requirement, by the way, so you can end up with adrenals or dual eswords with them.

Brasnak:
But that, in turn, would mean that maybe a lowpop antag balancing needs to be added in general.
The only question is, if it is going to be needed once the server launches or if there are enough players at all times that a second balancing is not needed.

FlattestGuitar:
Bird and I agreed that what imsxz said makes sense. Our rules already prohibit station desctruction and mass murder, so ideally people just won't do that. Since you can already buy a lot of strong shit and kill people, this shouldn't really change things much.

If this actually becomes a problem, please copy the link to this PR and PM it to me and call me a pancake. Thank you.

PR #139, opened 2017-10-17
Ports parabox

SamHPurp:
Another long slog port...

There will still be issues to resolve, but this does the majority of it.


AffectedArc07:
Wait a second. Where will tcomms go if its parabox?

SamHPurp:
@AffectedArc07 Its in mechanics spot atm, at the end of the hall above assembly.

code/game/area/Space_Station_13_areas.dm
```diff
@@ -382,6 +397,14 @@ NOTE: there are two lists of areas in the end of this file: centcom and station
...
 	icon_state = "hallC"
 
+/area/hallway/primary/central/ne
+
+/area/hallway/primary/central/nw
+
+/area/hallway/primary/central/se
+
+/area/hallway/primary/central/sw
+
```
AndrewMontagne: Is it worth giving these distinct names for when engineers have to go track down the relevant APCs?

FlattestGuitar: Not really something that /tg/ code does anywhere else. I'd like to see it happen, personally, but it's technically not a change this PR is about.

I'm fine with Purp adding it here or just making a **quick** PR after this one's merged to add the stuff to other maps.

SamHPurp: I like the unhelpfulness of "x has died in maintenance"

In the same vein of HELP MAINT...

Unuseful info

SamHPurp:
[Port Para Boxstation](https://trello.com/c/M37BuPkb/11-port-para-boxstation)

ktoma36:
Will we be keeping the engine containment the same or changing it for a SM shard? We could just change up the atmos containment areas and stick an SM shard containment area into where medical maint is now. Then just build maint around it!

taukausanake:
Maybe you can close up the containment area into part of Engineering and run the singulo inside. Then you just do whichever setup you want in the area. Sure the singulo will eat up the floor if done that way and maybe Lady Tesla will zap all things to hell if not grounded properly, but I'm sure something could be done that works

Xantholne:
Are we going for the updated para medical? I kinda like it more even if it's some 70% tg designed

Joseph-Case:
Why is this on hold may I ask?

SamHPurp:
We have so many other maps, and not enough pop for them already that I've de-prioritised this.

There are also so many significant changes in the codebases that I'm going to have to rework areas and how they flow... and I haven't had the appetite with all the other headminy things on that I need to be working on.

SamHPurp:
Gonna close, I don't have the time to finish this, and while we have 3-4 other maps working, this is superfluous to requirements atm.

PR #140, opened 2017-10-17
Oraclises Delta

SamHPurp:
Also fixes some misc issues


_maps/map_files/Deltastation/DeltaStation2.dmm
```diff
@@ -47549,6 +47555,7 @@
 	dir = 1;
 	layer = 2.9
 	},
+/obj/item/melee/chainofcommand,
```
FlattestGuitar: Of course it had to be Purpose who added that.

SamHPurp: 🍆

PR #143, opened 2017-10-18
Makes directional windows only block air if anchored

SamHPurp:
Porting this fix https://github.com/tgstation/tgstation/pull/31695

🆑 Purpose
fix: Directional windows now only block if they are anchored.
/ 🆑

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -306,7 +306,6 @@
 		return FALSE
 
 	setDir(target_dir)
-	air_update_turf(1)
 	ini_dir = dir
 	add_fingerprint(usr)
 	return TRUE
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/window.dm
```diff
@@ -207,6 +207,7 @@
 				to_chat(user, "<span class='notice'>You begin to [anchored ? "unscrew the window from":"screw the window to"] the floor...</span>")
 				if(do_after(user, decon_speed*I.toolspeed, target = src, extra_checks = CALLBACK(src, .proc/check_anchored, anchored)))
 					anchored = !anchored
+					air_update_turf(TRUE)
```
FlattestGuitar: I can't figure out why you only update air on the non-reinforced window after screwdrivering

code/game/objects/structures/window.dm
```diff
@@ -329,7 +330,6 @@
 		return FALSE
 
 	setDir(target_dir)
-	air_update_turf(1)
```
FlattestGuitar: You removed it from the counter-clockwise spin, but not the clockwise spin

FlattestGuitar:
ffs tg

SamHPurp:
@FlattestGuitar

PR #165, opened 2017-10-22
Fixes Plasma Fist disarming and grabbing

### LINKED ISSUE #155: Can't grab or disarm while using Plasma Fist martial arts
It literally returns basic hit for each one reeee.
I've got a fix in mind but I'm at school right now.
High priority, I enjoy using plasma fist while testing stuff on a private server.

imsxz:
:cl: imsxz
fix: Disarming and grabbing with Plasma Fist is now functional.
/:cl:

Fixes #155 
All went well in play testing the first commit, I don't think those last 2 commits here even did anything other than make the code look nicer.

### DIFF
--- a/code/datums/martial/plasma_fist.dm
+++ b/code/datums/martial/plasma_fist.dm
@@ -72,6 +72,7 @@
 /datum/martial_art/plasma_fist/disarm_act(mob/living/carbon/human/A, mob/living/carbon/human/D)
 	add_to_streak("D",D)
 	if(check_streak(A,D))
+		return 1
 		add_logs(A, D, "disarmed with Plasma Fist")
 	..()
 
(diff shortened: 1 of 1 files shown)


code/datums/martial/plasma_fist.dm
```diff
@@ -66,22 +66,22 @@
 	add_to_streak("H",D)
 	if(check_streak(A,D))
 		return 1
-	basic_hit(A,D)
-	return 1
+	add_logs(A, D, "punched with Plasma Fist")
```
FlattestGuitar: That will double up on the logs, though, no? You log it both here and in the parent function, unless I'm wrong.

imsxz: its how all the other martial arts do it, im just following precedent

PR #176, opened 2017-10-23
Removes mime and clown from Omega

SamHPurp:
Also discussed at length - if you've got a limited crew, where you place them really matters, an with all the lavaland roles going, these aren't a beneficial inclusion - ON OMEGA.

:cl: Purpose
del: The mime and clown are no longer shipped aboard Omegastation.
/ :cl:


imsxz:
:ok_hand: if we don't have most heads of staff on Omega, we definitely shouldn't have clown or mime.

taukausanake:
But clown and mime, clown especially, are paramount to SS13. If you removing them from any aspect then you're missing something important. Besides, who will have access to the theater if they don't?

PR #179, opened 2017-10-23
QOL change for Cargo

SamHPurp:
Makes flaps not pass atmos. This is how Paradise already handles this, and it has been frequently requested for Oracle.

:cl: Purpose
tweak: Plastic flaps no longer leak air
/ :cl:


Birdtalon:
Definitely needed.

PR #182, opened 2017-10-24
Stops being able to place tiles when someones buckled to a pipe on that tile.

SamHPurp:
Good god. What.

:cl: duncathan
fix: Can no longer tile tiles with the clown strapped to a pipe on it.
/ :cl:


FlattestGuitar:
Is this as hilarious as I think it is?

SamHPurp:
@FlattestGuitar yes

Birdtalon:
We should at least abuse this first 😄

PR #183, opened 2017-10-24
Fixes small lesser ash drake oversight

imsxz:
:cl: imsxz
fix: Fixes lesser ash drakes dropping loot when killed by kinetic crushers.
/:cl:

Megafauna drop special chests when killed with kinetic crushers. Lesser ash drakes aren't supposed to drop loot, as they aren't nearly as difficult to kill as a normal ash drake, but they were still able to drop crusher loot due to an oversight.

A neat exploit you could do with this is using the dragon shapeshift spell from lavaland, and letting a buddy kill you with a crusher over and over while reviving you somehow, for figuratively infinite ash drake chests.


FlattestGuitar:
Imsxz fixing exploits. I've seen it all.

PR #184, opened 2017-10-24
Removes Lust code from Sintouched. Adds in Hypocrisy sin instead. 

EldritchSigma:
Considering Lust wont be a thing heres a different one which is affecting alot more people IRL too. Should be a deadly sin aswell considering how severly it affects people.

:cl: EldritchSigma
balance: Removed deactivated code for Lust Sin as sintouched. Added Hypocrisy sin. Affects far more people IRL anyways than lust. And is equally as debilitating.
/:cl:

_Add tag: _Tiámâtuu diin nâmlúgal'lu; Why cant a Contributor erase their own PRs?_

### DIFF
--- a/code/game/gamemodes/miniantags/sintouched/objectives.dm
+++ b/code/game/gamemodes/miniantags/sintouched/objectives.dm
@@ -1,15 +1,8 @@
 /datum/objective/sintouched
 	completed = 1
 
-/*  NO ERP OBJECTIVE FOR YOU.
 /datum/objective/sintouched/lust
-
-/datum/objective/sintouched/lust/New()
-	var/mob/dead/D = pick(dead_mob_list)
-	if(prob(50) && D)
-		explanation_text = "You know that [D] has perished.... and you think [D] is kinda cute.  Make sure everyone knows how HOT [D]'s lifeless body is."
-	else
-		explanation_text = "Go get married, then immediately cheat on your new spouse." */
+	explanation_text = "Go get married, then immediately cheat on your new spouse by marrying an object... like a welding fuel tank." 
 
 /datum/objective/sintouched/gluttony
 	explanation_text = "Everything is so delicious. Go eat everything."
--- a/code/game/gamemodes/miniantags/sintouched/objectives.dm
+++ b/code/game/gamemodes/miniantags/sintouched/objectives.dm
@@ -1,8 +1,8 @@
 /datum/objective/sintouched
 	completed = 1
 
-/datum/objective/sintouched/lust
-	explanation_text = "Go get married, then immediately cheat on your new spouse by marrying an object... like a welding fuel tank." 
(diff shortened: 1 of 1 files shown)


taukausanake:
Whaaaaat? Isn't that just an RP objective then? I mean I guess it can be fun to force someone into a marriage and then slave them to yourself by it but I don't think that's how it'll be played

deep-fried-faceplant:
Fun alternative objective, get Ian a girl then kill him after they have a kid.

imsxz:
@taukausanake all of the sintouched objectives are RP I'm fairly sure.

SamHPurp:
I don't think this'll be a positive addition or encourage behaviors anyone actually wants.

deep-fried-faceplant:
Why not make the Lust objective something like, "Kidnap two crewmembers and force them to go on a date together". A dungeon in maintenance with some candles and wine would suffice.

taukausanake:
I guess I don't really see the appeal of sintouched objectives then. I would imagine there to be some function, like gluttony being able to physically eat all the food, or wrath at least having to cause harm because fuck people you hate them now.

There was this paste of a game I got a while back, pretty sure on Para Discord, and it was about a round somewhere else with traitors playing as sins. For Lust they were given "Make sure the heads don't leave the station and make sure they don't die". Maybe make it something similar to that, keep them trapped with you forever. Who knows, maybe they'll get Stockholm syndrome

SamHPurp:
@taukausanake have you played the Devil gamemode?..

Birdtalon:
No thanks. I'm all for having RP objectives but in game marriage is creepy and cringy and I'd rather not have an objective which tells people to do it built into our code.

EldritchSigma:
Altered. Other sin added instead that affects more people, abatu uruu,  usel'la mitu'ti
ikall-u bala'tut

deep-fried-faceplant:
I still like the idea of kidnapping people for the lust objective. Being bitchy and hypocritical seems like a waste of an objective slot for an antagonist.

taukausanake:
@SamHPurp No but I do believe things do need function. I understand the gamemode isn't for everyone except the devil player but function is still what I like. Giving them something to do, their own little endgame or like a mini antag; just not one that should cause damage

FlattestGuitar:
It's not even a real deadly sin. Get on with the times, heretics.

https://en.m.wikipedia.org/wiki/Seven_deadly_sins?wprov=sfla1

Birdtalon:
It's fine as an RP based objective. Doesn't really matter that it's not a deadly sin IMO. 👍

SamHPurp:
So I'm not keen on this or the alternative suggestions - personally.

SamHPurp:
If an alternate idea comes up, we're still open for it - but this isn't something we want as is.

PR #193, opened 2017-10-27
Balances Telebatons and Police Batons Stun/Cooldown Use Timers.

Xantholne:
Changes Telebatons and Police Batons to not have a 6 second stun time with a 4 second time before they can be used again, thus preventing them from being
1) Better than stun batons
2) Able to permately stun lock someone

:cl: Xantholne
balance: Telebatons and Police Batons have had their stun times and use times changed.
/:cl:

### DIFF
--- a/code/game/objects/items/melee/misc.dm
+++ b/code/game/objects/items/melee/misc.dm
@@ -130,7 +130,7 @@
 				if(check_martial_counter(H, user))
 					return
 			playsound(get_turf(src), 'sound/effects/woodhit.ogg', 75, 1, -1)
-			target.Knockdown(60)
+			target.Knockdown(40)
 			add_logs(user, target, "stunned", src)
 			src.add_fingerprint(user)
 			target.visible_message("<span class ='danger'>[user] has knocked down [target] with [src]!</span>", \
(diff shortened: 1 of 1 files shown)


FlattestGuitar:
A cautious yes, since I feel like the telebatons should be meant for self-defense, not assault squads.

Still, I'd rather we tested it properly first and came back to this PR in like two weeks to think about whether we want to keep it.

imsxz:
How else will I preform arrests as RD?!

deep-fried-faceplant:
Instead of just making a gap between stuns, could we do something where it takes a few hits in rapid succession to stun someone?

SamHPurp:
So the goal here is to make them defensive weapons, not offensive weapons - and better for utilisation of disablement, so you can get away to safety.

Makes sense to me, but might be worth discussion

Xantholne:
Yeah more so defensive than offensive, you can still drag someone out of your department while smacking them as a head, or if you so dare try to cuff then it'd be possible but highly unlikely. 

Four seconds _should_ be enough time to smack someone and drag them out the front door.The cooldown is just long enough where they'll get a chance to get up to resist, but it'd be better to run than try to stay in melee ranged at that point.

imsxz:
If we're willing to go with faceplants suggestion, stamia damage would do exactly that - require multiple hits to down someone. The only issue with stamina damage is it doesn't fix the issue of stunlocking once the target is downed.

Birdtalon:
Yes please.

PR #195, opened 2017-10-27
Makes radio colors the same as para

Joseph-Case:
Edited to make the radio colors make more sense (Like really, why is command a yellowy-greeny vomit color)

:cl: AffectedArc07
tweak: TComms colors are now more department accurate
/:cl:


FlattestGuitar:
Please also change the corresponding values in `interface/stylesheet.dm`

Joseph-Case:
@FlattestGuitar done

Joseph-Case:
image
An example of what the new colors look like

PR #196, opened 2017-10-27
Stops mice eating power cables

Joseph-Case:
Stops mice eating through wires.
While it does give the engineers something to do, It just isnt fun to go through maint finding **ONE** single eaten wire thats killing half the station

:cl: AffectedArc07
tweak: Mice no longer eat power cables
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/mouse.dm
+++ b/code/modules/mob/living/simple_animal/friendly/mouse.dm
@@ -60,21 +60,6 @@
 			playsound(src, 'sound/effects/mousesqueek.ogg', 100, 1)
 	..()
 
-/mob/living/simple_animal/mouse/handle_automated_action()
-	if(prob(chew_probability))
-		var/turf/open/floor/F = get_turf(src)
-		if(istype(F) && !F.intact)
-			var/obj/structure/cable/C = locate() in F
-			if(C && prob(15))
-				if(C.avail())
-					visible_message("<span class='warning'>[src] chews through the [C]. It's toast!</span>")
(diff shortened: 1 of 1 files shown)


FlattestGuitar:
Not sold on it, tbh. It's an interesting feature if it's rare enough and we don't have nearly enough time spent playing to know whether it's too high or too low.

I'd rather wait a bit longer, unless someone really wants to complain about this.

Alternatively, making mice eating cables print in dchat could be a good idea.

taukausanake:
Maybe have an alert on the power monitor that says connection to these APCs have been lost, please check for any broken power cables

SamHPurp:
I don't like this.

This is more of an issue due to low pop.

deep-fried-faceplant:
Have to agree with Sam and Flatty. If anything this might encourage engineers to learn the powergrid for the station.

Joseph-Case:
If it was only one station I wouldnt make this, but TG code has about 5 stations, can you really memorise the wire grid for every single one?

Birdtalon:
Hug Box. Them engineers need something to do!

SamHPurp:
@Joseph-Case don't memorise the grid - learn the mechanics and tools available to you for finding wire snips (multi-tools for example are super valuable!)

scrubmcnoob:
I like the factor of mice eating wires

FlattestGuitar:
Yeah, this isn't a significant issue that I can see so far. Engineers have quite a few different tools available to them that will greatly assist in solving the problems that arise from mice nomming wires.

We should wait for the playerbase to become more proficient at handling these things, since for now most people just don't know what they're doing. We should definitely revisit this if it becomes a larger issue.

PR #199, opened 2017-10-28
Point hotkey

scrubmcnoob:
straight to the point
:cl:
Middle mouse button now points.
/:cl:

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -274,8 +274,7 @@
 	return
 
 /*
-	Middle click out of sight
-	Only used for swapping hands
+	Middle click for pointing
 */
 /mob/proc/ShiftMiddleClickOn(atom/A)
 	return
@@ -456,8 +455,6 @@
 /obj/screen/click_catcher/Click(location, control, params)
 	var/list/modifiers = params2list(params)
 	if(modifiers["middle"] && iscarbon(usr))
-		var/mob/living/carbon/C = usr
-		C.swap_hand()
 	else
 		var/turf/T = params2turf(modifiers["screen-loc"], get_turf(usr))
 		params += "&catcher=1"
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -274,7 +274,7 @@
 	return
 
 /*
-	Shift Middle click
+	Middle click out of sight
 	Only used for swapping hands
 */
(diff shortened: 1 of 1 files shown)


code/_onclick/click.dm
```diff
@@ -274,20 +274,20 @@
...
+/mob/proc/ShiftMiddleClickOn(atom/A)
 	return
 
 /mob/living/carbon/MiddleClickOn(atom/A)
 	if(!src.stat && src.mind && src.mind.changeling && src.mind.changeling.chosen_sting && (iscarbon(A)) && (A != src))
 		next_click = world.time + 5
 		mind.changeling.chosen_sting.try_to_sting(src, A)
 	else
-		swap_hand()
+		..()
```
scrubmcnoob: So you don't randomly swap hands while you're trying to use a ling sting cause that's not supposed to be the button's function.

code/_onclick/click.dm
```diff
@@ -274,20 +274,20 @@
...
 
-/mob/living/simple_animal/drone/MiddleClickOn(atom/A)
+/mob/living/simple_animal/drone/ShiftMiddleClickOn(atom/A)
```
FlattestGuitar: This is only for drones. :v

code/_onclick/click.dm
```diff
@@ -368,7 +368,7 @@
 	A.CtrlShiftClick(src)
 	return
 
-/mob/proc/ShiftMiddleClickOn(atom/A)
+/mob/proc/MiddleClickOn(atom/A)
```
FlattestGuitar: There are no checks at all here. Dead people can point.

code/_onclick/click.dm
```diff
@@ -274,20 +274,20 @@
 	return
 
 /*
-	Middle click
+	Shift Middle click
```
FlattestGuitar: Shift click is not used for swapping hands in your implementation, from what I can tell.

code/_onclick/click.dm
SamHPurp: To be fair, ghosts can point on both TG and Para.

scrubmcnoob: It's not broken. Tested it. Alive people can't see ghost points.

code/_onclick/click.dm
scrubmcnoob: Fixed the note to correctly state what the function is now.

PR #202, opened 2017-10-29
Flightsuit Movement Adjustment one

EldritchSigma:
So flightsuits got the crazy TG movement speed apparently. Not sure if adjusting the Max Momentum Var will slow it down, but considering the boost speed var is setup as an acceleration it seems this would make sense.

Since we are on half of TGs movement speed I cut it in half for now. I am tired and will do more testing later on on this particular suit.

:cl: EldritchSigma
tweak: Adjusted Flightsuit Movement speed from TG movement speed to Oracle movement speed...
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/flightsuit.dm
+++ b/code/modules/clothing/spacesuits/flightsuit.dm
@@ -48,7 +48,7 @@
 
 	var/momentum_x = 0		//Realistic physics. No more "Instant stopping while barreling down a hallway at Mach 1".
 	var/momentum_y = 0
-	var/momentum_max = 500
+	var/momentum_max = 250
 	var/momentum_impact_coeff = 0.5	//At this speed you'll start coliding with people resulting in momentum loss and them being knocked back, but no injuries or knockdowns
 	var/momentum_impact_loss = 50
 	var/momentum_crash_coeff = 0.8	//At this speed if you hit a dense object, you will careen out of control, while that object will be knocked flying.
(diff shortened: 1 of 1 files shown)


SamHPurp:
Thanks for making this change. From our conversations about the equation being used - this all makes sense. Please compile and check it out when you can to confirm.

EldritchSigma:
Give it a shot. Seems pretty normal now.

code/modules/clothing/spacesuits/flightsuit.dm
```diff
@@ -48,7 +48,7 @@
 
 	var/momentum_x = 0		//Realistic physics. No more "Instant stopping while barreling down a hallway at Mach 1".
 	var/momentum_y = 0
-	var/momentum_max = 500
+	var/momentum_max = 250
 	var/momentum_impact_coeff = 0.5	//At this speed you'll start coliding with people resulting in momentum loss and them being knocked back, but no injuries or knockdowns
 	var/momentum_impact_loss = 50
```
imsxz: May want to halve this as well, otherwise you're losing 20% maximum momentum instead of 10%

imsxz:
Might wanna halve the momentum impact loss as well.

EldritchSigma:
Makes sense... halfed Momentum gain and impact lost aswell, I saw the boost mode earlier and was thinking it looked very off to accelerate like that. Impact lost so it doesnt feel like you loose like IMSXZ said 20% of your speed when hitting something when it was before only 10%.

SamHPurp:
You make a good point @imsxz

imsxz:
good first issue

PR #206, opened 2017-10-29
Plasma Cutter Nerf, lower rock pierce. Yes you can still pierce Liz- …

EldritchSigma:
…I mean ashwalkers with it.

So proposing this nerf to the plasmacutters because they are really breaking mining. They got an immense range and can pierce up to 5 rocks (the advanced variant). Even more if they are hitting diagonals. This should lower their efficiency while not making them completly useless. They are about as viable now as blast upgrades for the KA or the rarely used adv. Kinetic Resonator or the damage wise superior Kinetic Crusher and not just a straight upgrade from ALL the miners tools. Mech Cutter keeps its insane range and low rock pierce. If you dont have a super cell installed on your mech it drains the mech rather fast.  

:cl: Eldritch Sigma
balance: lowered plasmacutters and adv. plasmacutters mining pierce by a bit. Down to 1/3 of old range of the reg. one and 40% less pierce on the advanced. Still retain the 5 tile range. Still usefull for mining. As usual let me know if you still use it.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/special.dm
+++ b/code/modules/projectiles/projectile/special.dm
@@ -204,7 +204,7 @@
 	impact_effect_type = /obj/effect/temp_visual/impact_effect/purple_laser
 	var/pressure_decrease_active = FALSE
 	var/pressure_decrease = 0.25
-	var/mine_range = 3 //mines this many additional tiles of rock
+	var/mine_range = 1 //mines this many additional tiles of rock
 
 /obj/item/projectile/plasma/Initialize()
 	. = ..()
@@ -227,7 +227,7 @@
 /obj/item/projectile/plasma/adv
 	damage = 28
 	range = 5
-	mine_range = 5
+	mine_range = 3
 
 /obj/item/projectile/plasma/adv/mech
 	damage = 40


Birdtalon:
The changes look okay. I don't really have enough experience with mining to judge though.

SamHPurp:
Thats where I am with this... it seems fine to make the change - but I'm certainly not the best versed with this!

imsxz:
Ok so plasma cutter are definitely overpowered and they definitely do need nerfing.

Let's go over the pros and cons for your average miner (I'm on mobile so I can't do real bullet points RIP)

Pros:
-impossible to run out of ammo due to plasma ore being abundant for miners, each or sheet recharging half
-insane for miners, outclasses every other mining tool other than a KA with rapid repeater and mining explosion modkits
-its the best weapon against ash lizards, it takes off arms/legs in 2 shots on average

cons:
-not as good as KA vs standard fauna and megafauna

The biggest issue with them is that they're balanced around having limited ammo, but didn't consider the fact that the ammo is LITERALLY EVERYWHERE for a miner. 

In my opinion, this nerf is not the correct way to go about it. It's nerfing the stuff that the cutter is intentionally good at, and basically making it not worth choosing over the KA.

My proposed change is to remove plasma ORE recharging it, and making plasma sheets only charge a tiny bit (like 2-5% per sheet) or making it require charging via a recharger. This way it will still fill the role of being faster at mining than the KA, but not being infinite like the KA.

Xantholne:
Imsxz is very right with a lot of things. They're very good for mining in a straight/kinda diagonal line. They shoot about to the edge of the screen, maybe a bit more but also do shit for damage honestly against all types of Fauna; they cannot even hit a fauna unless it's within 3 tiles. If you fire it at a rock more than 2 tiles away it also does not penetrate making it shite. Personally I don't see the issue with them piercing so many rocks as provides an alternative to using KA with an AOE rock mining mod. The KIA can clear a lot of area very fast, but the plasma cutter can shoot further down the line if it penetrates early. In myy opinion going with a reduced charge from each piece of plasma sheets would be a good alternative. Otherwise I can't really see myself taking a cutter for mining than buying another KA and having one for mining and another for killing bads.

ktoma36:
I have to fully disagree with the current change as it essentially turns the advanced plasma cutter into a less useful KA with an explosive upgrade and makes the regular one plain useless. It is supposed to be an effective mining tool, the KA being much more effective at killing fauna with its explosive damage. If we were basing this solely off mining capability then yes, the advanced plasma cutter is the way to go to blast the most rocks per minute. The issue is though is that miners also have to stay alive while facing off dangerous wildlife/megafauna. While the plasma cutter rules mining it tickles when actually using it as a weapon to kill fauna. It takes about eight shots to kill a goliath with a plasma cutter and roughly fifteen seconds to fire those shots. An upgraded kinetic accelerator can kill a goliath in three to five shots and only take at most five seconds (If you spec for cooldown and damage). 

Im more for what imsxz is proposing. Don't nerf the unique advantage of the plasma cutters, instead make its drawbacks more prevalent. Decrease the amount of charge plasma ore and sheets give back to the cutter, right now a bit of ore will charge the thing something like 20-30% and a sheet a full 50% (I cant remember the actual amounts but it is rather insane). Perhaps dropping both raw ore and sheets down to 5-10% would work (Though sheets are actually more readily available once the ORM has been upgraded).

As for an ashwalker getting a hold of one... why does that apply at all to this change? 

As I said in the other nerf proposed, why make mining slower and more tedious when there are monsters to kill and treasure to loot? The plasma cutters are only prevalent cause they cut away the tedium of what is essentially clicking on rocks.

imsxz:
Currently, each ore recharges half of the cell. A plasma sheet will fully charge it.

SmokingKilz:
I agree with imsxz but it'd be nice to see its damage versus player controlled mobs to be reduced. Shooting off limbs with 1-2 shots is pretty dumb.

FlattestGuitar:
As per lengthy Discord discussion. The other PRs are getting merged, still.

PR #207, opened 2017-10-29
Plasmacutter tech and price increase

EldritchSigma:
While we are at it I looked at the price of the plasmacutters. They are really really low. Heck even tech requirements are dirt simple. You can print the regular ones even without the need of mining like 4 mins into the round. Not anymore. Considering they accelerate a bolt of plasma they now need a room temperature superconducting alloy made out of a diversity of transition metals (metal), uranium and silver. Well thats the fluff behind that change and why they should cost silver and uranium in the first place. The other one just needs a very very good conductor. Hence why it gets silver added as requirement. Oh and more techlevels so you cant hand the regular one out like cotton candy at roundstart since they are basicly better welders aswell.

:cl: EldritchSigma
balance: Increased techlevel requirements and resource requirements for the portable plasmacutters so they are more in line with the other mining tools.
/:cl:

### DIFF
--- a/code/modules/research/designs.dm
+++ b/code/modules/research/designs.dm
@@ -174,9 +174,9 @@ other types of metals and chemistry for reagents).
 	name = "Plasma Cutter"
 	desc = "You could use it to cut limbs off of xenos! Or, you know, mine stuff."
 	id = "plasmacutter"
-	req_tech = list("materials" = 3, "plasmatech" = 3, "magnets" = 2)
+	req_tech = list("materials" = 4, "plasmatech" = 4, "magnets" = 4)
 	build_type = PROTOLATHE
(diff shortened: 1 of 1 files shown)


ktoma36:
Why? I havent seen them being used by anyone other than mining in any of the test, let alone passed out like candy. They make mining faster, which keeps ores coming in and the station well supplied. Slowing down the R&D supply line will only make things take longer for seemingly no benefit. Also mining with a KA is tedious at best unless you get a mining aoe upgrade, which then decreases its combat capabilities. Making it harder to get plasma cutters and their advanced versions will hurt miners the worst.

Birdtalon:
With the change introduced in your other PR, do we really need to do both?

imsxz:
I don't agree with this change, see what I said in the last PR+what bird said.

PR #222, opened 2017-10-31
Port Many prs v2

scrubmcnoob:
ports:
tgstation/tgstation#31145
tgstation/tgstation#31165
tgstation/tgstation#31163
tgstation/tgstation#30297
tgstation/tgstation#29268
tgstation/tgstation#30633
tgstation/tgstation#31137
tgstation/tgstation#32162
tgstation/tgstation#30970
and yes it compiles.
these should be all the prs i ported, may of forgot to list the minor hotfixes ones i had to do
also did some other fixes for gangwars zlevel since we still have the game mode thankkfully


taukausanake:
Shouldn't you just be pulling from /tg/ upstream at this point? I mean I don't know the contents of these PRs/ports but I could only imagine such

scrubmcnoob:
We're too different code wise for a bot to handle @taukausanake 
A good amount of PRs would be scrapped. We don't want to look like FTL's git.

code/game/gamemodes/meteor/meteors.dm
```diff
@@ -31,17 +31,16 @@ GLOBAL_LIST_INIT(meteorsC, list(/obj/effect/meteor/dust)) //for space dust event
...
 		max_i--
 		if(max_i<=0)
 			return
 	var/Me = pickweight(meteortypes)
-	var/obj/effect/meteor/M = new Me(pickedstart)
+	var/obj/effect/meteor/M = new Me(pickedstart, pickedgoal)
 	M.dest = pickedgoal
-	M.z_original = ZLEVEL_STATION
-	spawn(0)
-		walk_towards(M, M.dest, 1)
```
FlattestGuitar: Lines 43 and 44 shouldn't have been deleted here.

SamHPurp:
Good catch, @FlattestGuitar 

I think I'd prefer if these ports remain their own individual PRs (other than subsequent fixes, those can all be rolled into one)

But if they're are entirely differnt, they should stay as separate PRs.

FlattestGuitar:
Yeah, I've already talked to Scrub about their approach.

Don't do this, kids.

FlattestGuitar:
Also, that comment I made earlier was apparently not quite right, since it works fine because Scrub also ported another PR which he forgot to mention in the OP.

PR #244, opened 2017-11-06
Removes Plasmamen, Pod People & Slime People as round-start races.

SamHPurp:
:cl: Purpose2
del: Removes Plasmamen, Pod People & Slime People as round-start races.
/:cl:

These aren't exactly interesting, or even particularly balanced at the moment. Removing this from being playable at round-start allow us to do more interesting things with them in things like adventure zones and admin events.

### DIFF
--- a/config/game_options.txt
+++ b/config/game_options.txt
@@ -400,10 +400,10 @@ ROUNDSTART_RACES human
 
 ## Races that are okay-ish on the station. Safe to enable.
 ROUNDSTART_RACES unathi
-ROUNDSTART_RACES plasmaman
-ROUNDSTART_RACES slime
+#ROUNDSTART_RACES plasmaman
+#ROUNDSTART_RACES slime
 ROUNDSTART_RACES ethari
-ROUNDSTART_RACES pod
+#ROUNDSTART_RACES pod
 
 ## Races that aren't on the station by default.
 #ROUNDSTART_RACES jelly


Cid777:
I'd personally like a completely clean slate, with just humans, to build off of. But anything that gets us closer to that is good.

Ziiro:
I'm against this, but I understand why you want to do it.

Personally, I think having a large selection races makes things more interesting and varied. Visually, mechanically, thematically, balance-wise. It adds a number of variables to a game that already has so many - I find this as a plus.

It also gives more roleplaying options, obviously. Granted, some people use species to give a sort of false depth to their character (Read as: the kind of people that always play their dwarves like Gimli in D&D )  but I prefer the option being there than not.

Balance wise I don't think the races slated for removal actually are imbalanced or bring anything extreme to the table, in the direction of powerful or weak.

taukausanake:
I doubt anything useful will be done with them as admin events. Maybe adventure zones but I still wouldn't bet on it. I mean just look at Paradise with their Wryn and Nucleations. Bad comparison with the two codebases but my point stands

Cid777:
> Balance wise I don't think the races slated for removal actually are imbalanced or bring anything extreme to the table, in the direction of powerful or weak.

I mean that's the problem with them, isn't it? They're pretty uninteresting.

> Personally, I think having a large selection races makes things more interesting and varied.

I actually agree. But I think it makes sense to remove the uninteresting ones _while we can_. Better ones could fill the void later, there's no shortage of people wanting new races.

> Bad comparison with the two codebases but my point stands

I apologize if I come off as rude, but your point _doesn't_ stand when you immediately undermine it.

Ziiro:
>I mean that's the problem with them, isn't it? They're pretty uninteresting.

The problem is with SS13: How do you do interesting without it being overpowered or underpowered? It's a very, very difficult line to walk. I think if something is a little bland, but functional, that's okay.

imsxz:
More room for moths!

imsxz:
Also how will this affect people who already have a character of the to-be-removed race?

FlattestGuitar:
I disagree with this PR myself. We should avoid to have boring unplayed species like Drask or Kidan - barely anyone picks those on Paradise and they're just bloat. 

All the species listed here are actually played. Fetish or not, people only pick these things because they *want* to.

Taking away the ability to pick those barely impacts balance and just strips the game of what interesting interactions could take place as a result of how the two species interact with the world.

:-1:

Cid777:
> We should avoid to have boring unplayed species like Drask or Kidan - barely anyone picks those on Paradise and they're just bloat.

I agree, the sooner you remove bloat the easier it is.

> All the species listed here are actually played. Fetish or not, people only pick these things because they want to.

Technically true, but misleading. Pod people aren't really played, people pick them once for a test then never again.

> Taking away the ability to pick those barely impacts balance and just strips the game of what interesting interactions could take place as a result of how the two species interact with the world.

That's a bit dramatic, don't you think? And I feel it needs to be stressed, it's not like races are going to be cut out and _that's it_, at least I don't think so. Races can, and probably will, be added in their place over time.

Birdtalon:
Plasma and Pod people sure. But lots of people play slimes. Is that really a good idea to remove that?

Brasnak:
From a pure mechanical view, plasma and slime people are the most distinct round-start races, in comparison to humans, we have right now.

Xantholne:
I don't even think we have pod players. We do have a few Plasmaman players though so I'd hate to remove their characters.

Slime wise we have quite a few of those and would rather they be kept around to still keep some race variety.

SomeGuy9283:
Let's do it!

Humanity First!

imsxz:
We're still in the play testing phase, it would be best to do something like this before the full release. ~~remove dogbeasts too~~

SamHPurp:
Worth me stating - we're not against re-introducing these in the future. But it'll be when they're made much more interesting / stricter criterium

PR #246, opened 2017-11-06
Updates the Ion Law string set

SamHPurp:
Removes some things that would promote behaviours we probably shouldn't include.

I've also added a few Oracley things.

Please have a review let me know if you think I've gone too far, or not far enough.


strings/ion_laws.json
```diff
@@ -535,10 +521,10 @@
         "PRESS B",
         "SMELL LIKE THE MAN YOUR MAN COULD SMELL LIKE",
         "PIRATE VIDEO GAMES",
-        "WATCH PORNOGRAPHY",
         "INSULT THE LIZARDS",
         "FLIRT WITH THE LIZARDS",
-        "GAS THE LIZARDS"
```
imsxz: I personally feel we should promote this type of behavior :thinking:

SamHPurp: Slaughter, Sacrifice, Maim... sure.

Gas, is a bit too... Final Solution for me.

strings/ion_laws.json
```diff
@@ -535,10 +521,10 @@
...
+		"INSULT THE ETHARI",
+		"FLIRT WITH THE ETHARI"
```
taukausanake: :eyes:

strings/ion_laws.json
```diff
@@ -943,8 +925,6 @@
         "PETES",
         "RATVARIANS",
         "XENOS",
-        "FETISHES",
-        "MEMES",
```
taukausanake: Muh memes! Don't take away muh memes!

strings/ion_laws.json
SamHPurp: I wasn't going to add it, but it existed for lizards. Yeah those could both go, I'd be more than happy with that.

strings/ion_laws.json
SamHPurp: We all know AIs will be memeing anyway, regardless of laws @taukausanake

strings/ion_laws.json
taukausanake: I'm not saying remove it, I'm just surprised you put it there

SamHPurp: *I am too*

FlattestGuitar: 🍆

PR #249, opened 2017-11-06
nerfs flightsuit zoom

SamHPurp:
ports https://github.com/tgstation/tgstation/pull/32312

totally should never have been 14. 12 is much saner.


FlattestGuitar:
Can't argue with that.

PR #252, opened 2017-11-06
Fixes directional character previews

SamHPurp:
Ports https://github.com/tgstation/tgstation/pull/32332 from upstream

:cl: deathride58
fix: Directional character icon previews now function properly. Other things relying on getflaticon probably work with directional icons again, as well.
/:cl:


FlattestGuitar:
Now I can check out my Ethari's bottom. 10/10

PR #271, opened 2017-11-10
Reworked Floor Tiles

AndrewMontagne:
This PR completely reworks floor tiles. They look new and clean, so they can be made dirty properly. The reinforced plating and inspiration was taken from VOREstation.

image
(click to enlarge)

Also, this PR fixed damage to plating/tiles not respecting the colour of the tile.

image

:cl: AndrewMontagne
fix: Damaging floor tiles now keeps the styling.
imageadd: Reworked floor tile sprites
/:cl:


AndrewMontagne:
This PR was so tedious to implement. D:

code/game/turfs/simulated/floor.dm
```diff
@@ -87,6 +88,9 @@
 
 /turf/open/floor/proc/update_icon()
 	update_visuals()
+	cut_overlays()
```
FlattestGuitar: `update_visuals()` is responsible for atmos overlays, like the icy stuff we added before. Calling `cut_overlays()` right after will remove that. No bueno.

Will be best if you do `cut_overlay(current_overlay)` instead, though I'm actually not sure if that will work. Do try!

AndrewMontagne:
This was merged prematurely, PR will be remade.

PR #283, opened 2017-11-10
Adds a button on the UI for the Web-Map

Joseph-Case:
Throws a button for the web-map right in peoples faces because ~~I want my work seen~~ this is a useful tool. Ignore the 7 commits it took for me to make this.

:cl: AffectedArc07
add: The UI now has a dedicated button for the webmap.
/:cl:

I feel this would be a good addition as its a useful tool for learning the map

### DIFF
--- a/interface/skin.dmf
+++ b/interface/skin.dmf
@@ -904,8 +904,8 @@ window "infowindow"
 		type = BUTTON
 		pos = 329,5
 		size = 38x20
-		anchor1 = 51,1
-		anchor2 = 57,5
+		anchor1 = 51,0
+		anchor2 = 64,0
 		background-color = none
 		saved-params = "is-checked"
 		text = "Web Map"
--- a/interface/skin.dmf
+++ b/interface/skin.dmf
@@ -1,778 +1,778 @@
 macro "default"
-	elem
+	elem 
 		name = "TAB"
 		command = ".winset \"mainwindow.macro=hotkeys mapwindow.map.focus=true input.background-color=#e0e0e0\""
-	elem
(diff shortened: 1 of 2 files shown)


SamHPurp:
Image of it please?

interface/skin.dmf
```diff
@@ -929,26 +949,16 @@ window "infowindow"
...
 		pos = 96,5
-		size = 88x20
+		size = 86x20
 		anchor1 = 15,0
 		anchor2 = 29,0
 		background-color = none
 		saved-params = "is-checked"
 		text = "Rules"
 		command = "rules"
-	elem "wiki"
```
FlattestGuitar: You removed the wiki button. Don't remove the wiki button.

Joseph-Case:
I will get one when I am home

Joseph-Case:
image

FlattestGuitar:
Too small on my 16:10. Make it bigger.

Joseph-Case:
Can I just change the text to "Map" and leave it at that?

Joseph-Case:
I will attempt to make them all the same size at some point

Birdtalon:
The only reason I don't like this is because it links to somewhere outside of the sphere of control and therefore we cannot guarantee that it's necessarily a trusted site. The web-map would have to be brought in-house to our own infrastructure for me to be comfortable with it.

Joseph-Case:
Ok, The code is available on the github fully working after dinking around with it today, I will close this for now, when you want me to add the code to its own web github, just ask and I will PR it over.

PR #289, opened 2017-11-11
Fixes traitor objectives for non-existent items

AndrewMontagne:
image

:cl: AndrewMontagne
fix: You can no longer get a steal objective for a non-existent item.
/:cl:


SamHPurp:
I don't think this works like this.

The supermatter slither objective for example will never proc... blueprints will proc even if there is no CE locker because they spawn in some Lavaland areas...

Another that won't proc is the filled plasma tank?

I think this'll be slowing down load times & not actual achieve the desired effect.

AndrewMontagne:
@SamHPurp Fair cop on the items that don't exist at roundstart, or exist on other z levels. As for the performance impact? It's honestly minimal. It'll only be invoked when enumerating objectives, which is only once per round.

PR #304, opened 2017-11-17
Drape murder

FlattestGuitar:
Drapes are dead.

Use any sharp object on help intent while your patient is laying down to start a surgery.

You still need to do everything you had to do before to properly operate.


Ziiro:
the absolute mad man, he actually did it

PR #307, opened 2017-11-17
*flip 2.0

FlattestGuitar:
Ports some *flip stuff from Gooncode, for fun.

Most importantly, you can now flip over people you've grabbed. This is a very interesting thing to use in combat, but I'm not sure if everyone agrees we want it. I personally love it.

### DIFF
--- a/code/modules/mob/living/emote.dm
+++ b/code/modules/mob/living/emote.dm
@@ -140,9 +140,59 @@
 	message = "does a flip!"
 	restraint_check = TRUE
 
+/datum/emote/living/flip/can_run_emote(mob/user, help_check)
+	if(..(user, help_check))
+		if(!user.incapacitated())
+			return TRUE
+	return FALSE
+
 /datum/emote/living/flip/run_emote(mob/user, params)
-	if(..())
+	//the emote doesn't need much magic, so I'm overwriting everything. Bite me.
+	if(!can_run_emote(user))
+		return FALSE
+
+	if(istype(user.loc, /obj))
+		var/obj/container = user.loc
+		to_chat(user, "<span class='warning'>You flip and smack your face into [container]!</span>")
+		container.visible_message("<span class='warning'><b>[container]</b> emits a loud thump and rattles a bit.")
+		if(isliving(user))
+			var/mob/living/L = user
+			L.Knockdown(100)//10 seconds
+		playsound(user.loc, "sound/effects/bang.ogg", 50, 1)
+		var/original_x = container.pixel_x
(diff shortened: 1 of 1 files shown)


SamHPurp:
I don't want this. Hell I don't even want a flip emote.

FlattestGuitar:
It's okay to be wrong, Purpose.

EldritchSigma:
By flipping people in combat you mean suplexing them? That would be very meth. Shitty 1.20 meters Spacebird suplexes 2 meters tall dustlung to death.

Birdtalon:
Flipping is great and adds a nice degree of light hearted fun.

FlattestGuitar:
To clarify, when you flip while holding someone you two just switch places and the grab is lost. No stuns or knockdowns or anything similar.

code/modules/mob/living/emote.dm
```diff
@@ -140,10 +140,60 @@
...
+		container.pixel_y = original_y
+		if(prob(2) && istype(container, /obj/structure/closet))
+			var/obj/structure/closet/C = container
+			if(C.locked)
+				C.bust_open()
+
+	else if(user.IsKnockdown() || user.lying)
+		user.visible_message("<b>[user]</b> flops and flails on the floor!")
+		//don't do anything! D:
+	else if(user.pulling && isliving(user.pulling))
```
Fox-McCloud: You'll want to prevent this from working if either person is buckled; that's not checked for in `incapacitated` last I knew.

code/modules/mob/living/emote.dm
```diff
@@ -140,10 +140,60 @@
...
+		var/mob/living/L = user.pulling
+		var/turf/tmp = get_turf(L)
+		var/turf/T = get_turf(user)
+		if(tmp && T)
```
Fox-McCloud: This is going to lead to all kinds of messed up things. 

Like someone pulling someone, getting inside something, then flipping and them moving out of it.

You need to check that the target and destination are both turfs.

ktoma36:
God I would love/be terrified if we could get the wrestler's belt from goon, it is terrifying to watch one man with a folding chair wrestle a station to death

Xantholne:
We have the wrestling belt. Not sure if it is goon's but you can in fact wrestle with it. Works like krav maga gloves

Fox-McCloud:
It's based on Goons, but it's kinda a quick and dirty implementation of it.

SamHPurp:
I guess I'll have to concede on this for now. -_-

PR #311, opened 2017-11-17
Added a trash cart, disposals to Omega.

AndrewMontagne:
:cl: AndrewMontagne
add: Adds a trash cart and disposals route to Omega.

/:cl:


Joseph-Case:
Does it have disposals shipping?

PR #314, opened 2017-11-18
PDA slot > ID slot

FlattestGuitar:
It's just better, okay.

When you press E the PDA slot will be prioritized. The only consequence is putting a PDA on the PDA slot instead of the ID slot. o7o7o7o7o7


SamHPurp:
was about to go nuts at the title.

Yeah thats fine.

PR #317, opened 2017-11-21
Ports Paradise SecHUD Icons

Joseph-Case:
Ports the paradise HUD icons, as I generally find they are nicer than the TG ones.
Examples:
Assistant:
image

Head of Security:
image

All Status Icons (Sec hud, medical hud, tele+chem implants, loyalty):
image

:cl: AffectedArc07
imageadd: Added Paradise HUD Icons
/:cl:


AndrewMontagne:
@FlattestGuitar @SamHPurp Can we get review on these please?

SamHPurp:
I guess I'm not against it, but I kinda like TGs.

Joseph-Case:
@SamHPurp You find the TG mindshield/loyalty thing better? I can barely tell when people are shielded, especially when all the roles that are meant to have mindshields have mainly blue icons

SamHPurp:
Can you provide screenshots of TGs so we can side-by-side them?

I just find the giant green Paradise dot ugly as fuck.

Joseph-Case:
@SamHPurp will get one, Also the captain icon is only a [?] because I had the ID in the sec records when I was adding a wanted status, I cant get a picture of it with an ID in if needed, as for the picture here it is:
image
Literally the only way to tell if they are mindshielded/loyalty implanted is by a 1 pixel blue ring around the badge

FlattestGuitar:
Personally, I feel there are many icons in both sets that are *shit*, but in the end I really like /tg/'s version more. I think we have some room to do something pretty and more modern than a giant green blob. A pretty animation on a dude would be incredible. For now, I'd rather we stuck with /tg/'s than switched to Para's.

Sorry for taking so long to respond.

SamHPurp:
Thats essentially where I am with it too..

I agree with you @Joseph-Case that it could be better. I just think Paradise's green dot solution is worse, rather than better.

Something we could develop/design ourselves rather than just port perhaps?

Joseph-Case:
Ok, I see your point.
I will try to think of a better way to identify mindshields

Joseph-Case:
If I could figure out directional sprites, mindshielding could be done the same way as that genetic effect which gives an orange glow around the sprite

SamHPurp:
As we discussed, we agree it could be done better, but not really good with Para's way.

Entirely encourage you to submit a new PR with a new way of identifying this!

Spartan6:
*scream

PR #321, opened 2017-11-25
".w" whisper and different chat disable message

Brasnak:
Adds the ability to whisper by using ".w" as a prefix during Say. Also changes the chat disabled message from "Speech is currently admin-disabled" to "Speech has currently been disabled by the admins."


SamHPurp:
Leading merge commits againnnnnnnn

code/modules/mob/say.dm
```diff
@@ -2,17 +2,21 @@
 /mob/verb/say_verb(message as text)
 	set name = "Say"
 	set category = "IC"
+	var/regex/whisper_finder = regex("^\\.w\\s+|^\\.w")
 	if(GLOB.say_disabled)	//This is here to try to identify lag problems
-		to_chat(usr, "<span class='danger'>Speech is currently admin-disabled.</span>")
+		to_chat(usr, "<span class='danger'>Speech has currently been disabled by the admins.</span>")
```
FlattestGuitar: This is actually grammatically incorrect, meine Freunde.

FlattestGuitar: Or would it be "mein Freund"..?

I always get it mixed up.

SamHPurp:
Can we be super fucking careful adding more regex.... They still haven't fixed the Regex that Flattest added to Paradise which stops the server from running nine times out of ten.

code/modules/mob/say.dm
Brasnak: "mein Freund" is singular, "meine Freunde" is plural.

Brasnak:
i did test it, and so far, i have seen it cause no trouble at all

code/modules/mob/say.dm
Brasnak: Also, could you point out the grammatical error to me? I am failing to spot it.

FlattestGuitar:
a) This regex is fine, it has no look-behind stuff, which apparently breaks BYOND on Linux (that's how I sabotaged Paradise back in the day)

b) `Speech is currently admin-disabled.` is correct. `Speech has recently been admin-disabled.` is also correct. `Speech has currently been admin-disabled.` is not correct. Leave it the way it was :P

Brasnak:
Change reverted

PR #326, opened 2017-11-29
Travis CI Artifacts & Shutdown For Update

AndrewMontagne:
This adds a step to the CICD pipeline which uploads the artifacts to an S3 bucket if they're not a pull request. It also adds a config option to kill the server instead of rebooting it when World.Reboot() is called if the server needs updating, so that the supervising daemon can update the server and restart it.

No changelog as it has no visible changes.


tools/CreditsTool/UpdateCreditsDMI.sh
```diff
@@ -1,13 +1,13 @@
...
-
-rm ../../icons/credits.dmi
-
-for filename in credit_pngs/*.png; do
-	realname=$(basename "$filename")
-	java -jar ../dmitool/dmitool.jar import ../../icons/credits.dmi "${realname%.*}" "$filename"
-done
-
-rm -rf credit_pngs
+#!/bin/bash
```
AndrewMontagne: The line endings on this file were incorrect, and have been fixed.

SamHPurp:
Above my paygrade to review this one. @Allfd & @FlattestGuitar ?

PR #342, opened 2017-12-05
Advanced E-Gun has a slight malfunction chance again.

EldritchSigma:
Since nerfing the firerate would nerf the E-Guns firerate aswell and just increasing the recharge time would basicly just changing one number without any other interactions other "Duh carry another" I propose to reactivate the irradiation and EMP fail chance on the E-Gun. 

What it is supposed to do is to have fail ticks. Each time you use the gun or when it recharges itself (or changes its state in any way, yes this should include EMPs) it has a chance to run into a fail tick. If a fail tick happens a few things can happen. It can irradiate the user, requiring the user to get some Potassium Iodide. It can strongly irradiate the user which might result in more severe things than just a few radiation burns. On getting hit by an EMP it runs the same failchecks but at a much higher likehood to pass.

The reason for this nerf is that the AEG currently has an insane firerate, fast recharge and a cheap price which basicly outclasses everything sec can get short of ballistics for the occasional esword using prick. Sec will visit the chemist in medbay and ask for Anti-Rad pills containing charcoal, saline glucose and of course Potassium Iodide to heal the burns. Or you know... actually wear a damn hardsuit or heavier armor from the armory if they intend on using this weapon mainly. It is an experimental military grade weapon and should not be used without any good reason. That should stop Captains like that fucking Thali on getting one every other round because it just outclasses the E-Gun and even the Antique Laser by far at the moment.

:cl: EldritchSigma
tweak: The AEG now irradiates you again if you use it overextensivly.
experiment: Nanotrasen Weapon Developement Department in collaboration with Omegacorp Arms found out that the new Advanced Energy Gun, commonly known as AEG, has a severe flaw in its design. The Thorium Breeder Reactor used to make it a self sufficient weapon cannot be shielded efficiently without making this weapon alot bulkier and impracticable. Nanotrasen recommends that any personell intending to use it should inform themself about prolonged exposure to irradiation and how to treat radiation exposure. 
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/energy_gun.dm
+++ b/code/modules/projectiles/guns/energy/energy_gun.dm
@@ -130,11 +130,11 @@
 		switch(fail_tick)
 			if(0 to 200)
 				fail_tick += (2*(fail_chance))
-				M.rad_act(20)
+				M.rad_act(40)
 				to_chat(M, "<span class='userdanger'>Your [name] emits alot of heat...</span>")
 			if(201 to INFINITY)
 				SSobj.processing.Remove(src)
-				M.rad_act(40)
+				M.rad_act(60)
 				crit_fail = 1
 				to_chat(M, "<span class='userdanger'>Your [name]'s micro-breeder reactor overloads!</span>")
 
--- a/code/modules/projectiles/guns/energy/energy_gun.dm
+++ b/code/modules/projectiles/guns/energy/energy_gun.dm
@@ -134,7 +134,7 @@
 				to_chat(M, "<span class='userdanger'>Your [name] emits alot of heat...</span>")
 			if(201 to INFINITY)
 				SSobj.processing.Remove(src)
-				M.rad_act(60)
+				M.rad_act(80)
 				crit_fail = 1
 				to_chat(M, "<span class='userdanger'>Your [name]'s micro-breeder reactor overloads!</span>")
 
--- a/code/modules/projectiles/guns/energy/energy_gun.dm
+++ b/code/modules/projectiles/guns/energy/energy_gun.dm
@@ -111,8 +111,8 @@
 	ammo_x_offset = 1
 	ammo_type = list(/obj/item/ammo_casing/energy/electrode, /obj/item/ammo_casing/energy/laser, /obj/item/ammo_casing/energy/disabler)
 	selfcharge = 1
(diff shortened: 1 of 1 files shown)


FlattestGuitar:
I think I'm okay with this, but I'd like the AEGs removed and replaced with something more interesting anyway.

imsxz:
I don't see the purpose in nerfing it. The whole idea around science is providing general better version of currently existing shit, attempting to leave it on par with the other energy gun really defeats the whole idea.

The main issue with sci guns is how easy they are to produce in mass. I had an idea for a solution to this, where RND would be able to print upgrades to already existing guns, such as a self charging cell upgrade for energy based weaponry, requiring cargo cooperation to get lots of super strong guns instead of sci just printing shitloads.

EldritchSigma:
Dehugboxed the irradiation. Was definitly too low on magnitude. Think the initial values are just fine. Expected them to mutate you wildly. Which is not the case.

Anyways RnD: Should keep the guns in my opinion... but rather give out unstable prototypes and then , once a prototype has been delivered to centcomm, cargo can order the finished product without malfunctions/deficits. The AEG currently OUTPERFORMS the unique gun of the HoS, the Laser of the Captain and pretty much anything Sec and Antags can get because it fills any role without any drawback (short of not having a 3D printer to shoot stetchkin bullets). Only thing it is inferior to are indeed Pulse weapons, but only because these have a high ammo capacity and a better lethal mode which is good. It isnt much fun for the antag to get permastunned by one of those things because a tazer bolt just stuns long enough for the gun to recharge a shot and then fire another one (or get sprayed by tazer shots because of its insane firerate), nor is it much fun for sec to get hit by this monstrosity either. Hence why it should have a weakness and the malfunction is an interesting mechanic (it heats up and requires moderation in use) that was already present without "just lowering the recharge rate so it cannot be used permanently".

SamHPurp:
So I totally agree, we should do something to weaken weaponry somewhat.

I don't think this is the right way to do so though. We had a bit of a debate on it last night on the #coder-stuff channel, and I think we all agreed to that point at least.

Entirely out of scope of this, but a larger re-work (perhaps encompassing how RND works) would be beneficial instead.

SamHPurp:
Agreed regarding the RND printing, I think that _might_ be the sort of system we should move towards. However out of scope of this.

PR #347, opened 2017-12-05
Removes Hulk

Ziiro:
Stun immunity in a combat system based around stuns is a bad idea.

(thanks goof https://github.com/tgstation/tgstation/pull/31576 )

### DIFF
--- a/code/datums/mutations.dm
+++ b/code/datums/mutations.dm
@@ -628,8 +628,7 @@ GLOBAL_LIST_EMPTY(mutations_list)
 /datum/mutation/human/laser_eyes
 	name = "Laser Eyes"
 	quality = POSITIVE
-	get_chance = 15
-	lowest_value = 256 * 12
+	dna_block = NON_SCANNABLE
 	text_gain_indication = "<span class='notice'>You feel pressure building up behind your eyes.</span>"
 	layer_used = FRONT_MUTATIONS_LAYER
 	limb_req = "head"
--- a/code/datums/mutations.dm
+++ b/code/datums/mutations.dm
@@ -116,9 +116,8 @@ GLOBAL_LIST_EMPTY(mutations_list)
 /datum/mutation/human/hulk
 
 	name = "Hulk"
+	dna_block = NON_SCANNABLE
 	quality = POSITIVE
-	get_chance = 15
-	lowest_value = 256 * 12
 	text_gain_indication = "<span class='notice'>Your muscles hurt!</span>"
 	species_allowed = list("human") //no skeleton/lizard hulk
 	health_req = 25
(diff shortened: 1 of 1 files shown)


FlattestGuitar:
I like this tbh.

Ziiro:
The one part I'm unsure of is Laser Eyes are quite strong. I was looking into trying to make balance adjustments (for example, nutrition cost). However I can't seem to track down where those values are stored.

Xantholne:
This is much better IMO than hulk.

SamHPurp:
Removing hulk 🆗 

Laser eyes though... its super fukken strong....

imsxz:
Ziiro, based on past conversations, I think you might have a definitions problem, in this case what balance means. What I think you need to work on, is this idea that outright removing things that seem to be unbalanced or things that seem to not work towards balance does mean you are making the better. I think this is a really good example, because as we can see in this example you clearly are frustrated at the current state of hulk and the power it gives others and the ease at which the power is received, but you acknowledge that just simply removing it would leave players discontent, so you vouch to replace it with a power that, coincidentally, is equally if not more powerful. What I believe to be the problem, and I've seen this in other situations which you bring up, is that a lot of times you have some valid complaints or realize something that has a significant degree of trueness to it, but then use this to justify handling the situation in an extreme or otherwise unneeded way that while possibly justifiable, doesn't actually solve the problem you brought up, or at least not without creating new problems. How to solve this? Not sure, and I think it depends on the situation. Maybe in this case, instead of out right removing things, create a more interesting and challenging approach to obtaining more powerful powers. This would not only solve the problem, but also create benefits that didn't previously exist, in this case by providing an additional challenge for a job that's usually over in 10-15 minutes and adding new cool things to achieve for (in this case hulk and the previously unobtainable powers) which now adds new life to an aspect of the game that has been steadily growing mundane. Just an idea and something to think about.

I've really liked many of your ideas in the past and I'd hate to see certain nit-picky things become your downfall. In the end it comes down to the mindset of remembering that while pruning a tree can help its growth, if you keep chopping it up with out fertilizing it a bit as you go along, soon enough you be left without a real tree.

imsxz:
Copy pasta aside, pretty sure laser eyes have some balance changes compared to paradises ridiculous laser eyes. iirc they drain hungry, but fukken virology mitigates this entirely with weight even.

it's likely worth noting that on TG, validhunting is generally unrestricted and hulks are much more relevant due to this. In a server where validhunting is generally not allowed, hulks will only shine when AI is malf or xenos are on station. Meanwhile on TG, you show your face with syndie gear equipped and the geneticists are quite literally bashing down your doors looking for VALIDS

SamHPurp:
Lets maybe limit shitposting on the Git for now. 👀

Ziiro:
Even if the culture and the game we play is different, the idea behind hulk is terrible from a balance perspective. Everywhere else in the code, anti-stun is handled with pretty careful measure. Except for hulk. Randomly obtained from genetics, free to give to as many people as you want, and potentially hard to take away in the hands of a good player.

If the wall busting is really that important, we could move a nerfed version of it over to a new power.

And if people really, *really* think we need Hulk, I could change the health threshold at which it disables to something more reasonable. (Like `if(owner.health < 40)`

Ziiro:
Per discussion, laser eyes removed too

NO HULK
NO LASER EYES
FINAL DESTINATION

EldritchSigma:
Genetics also can get all powers without any drawbacks. I like for once what para did with the genetic instability that requires you to balance it out with disabilities. Also para got a far wider selection of powers, rivavling the ammounts of powers you can get from goon

FlattestGuitar:
I am fine with this. As per our conversation, 🤤

PR #351, opened 2017-12-06
Backpack Storage Changes

DaveTheHeadcrab:
:cl: Dave
tweak: Backpacks may no longer be accessed while they are equipped.
tweak: Satchels may be accessed while equipped, but the storage capacity of satchels has been reduced.
/:cl:

All storage items have a new var named very verbosely: block_open_while_equipped. If the items have slot_flags and this var is set, the storage item will fail to open if it is found within the contents of the opening mob's get_all_slots() list.

Backpacks and duffel bags have block_open_while_equipped set.

Satchels do not have block_open_while_equipped, however their storage capacity has been altered from the standard max_combined_w_class and storage_slots value of 21 to a new value of 9. This means that a satchel could fit 3 'normal' sized items in it, and up to 9 'tiny' items in it.

Code is untested, but it's too late for me to do it tonight. Will test it soon.
Do not even think of merging this before it's tested or else your family will get an ancient chinese curse and probably start bleeding from the eyes.

**EDIT:** As requested, this PR now moves get_all_slots() further up the inheritance chain and adds a new override for mobs of type `/mob/living/carbon`. Still very untested.


Ziiro:
I understand where you're coming from on this and it is interesting, but I don't necessarily agree that it's an entirely positive change.

I'd be willing to try it on the grounds of it being a test/trial. That said, I feel like the end result is just going to be annoyance and tedium of having to click your backpack off, and then use your free hand to rifle around in it. This is also a pretty fundamental change in the way SS13 works that might put people off.

Not to mention we'd probably need a *lot* more lockable lockers in departments for people to secure their things. Perhaps a system to name the locker.

The balance implications are potentially weird as well. On the plus side, less walking armories and have to prepare more carefully. On the other, it'll feel shitty how limiting it is when you're something like a sec officer that needs a lot of equipment. I think we'll need to see an expansion of exosuit/outterwear clothing items that have pockets.

imsxz:
I like the idea of this. Perhaps for a QOL thing, we could keep the ability to view the contents of your backpack without taking it off? I believe this is a good change because it's super unrealistic to pull shit out of your backpack with ease while it's still on your back.

This change is similar to the whole "no eating while mouth is covered thing" in that it may seem super tedious at first, but it won't be that difficult to adapt to in reality.

Spartan6:
While super interesting, this feels... a bit over the top.

code/game/objects/items/storage/backpack.dm
```diff
@@ -197,6 +199,9 @@
 	desc = "A trendy looking satchel."
 	icon_state = "satchel-norm"
 	species_exception = list(/datum/species/angel) //satchels can be equipped since they are on the side, not back
+	block_open_while_equipped = 1
```
Ziiro: Wait shouldn't this be `block_open_while_equipped = 0` so it _can_ be opened?

code/game/objects/items/storage/storage.dm
```diff
@@ -25,7 +25,7 @@
 	var/collection_mode = 1;  //0 = pick one at a time, 1 = pick all on tile, 2 = pick all of a type
 	var/preposition = "in" // You put things 'in' a bag, but trays need 'on'.
 	var/rustle_jimmies = TRUE	//Play the rustle sound on insertion
-
+	var/block_open_while_equipped = 0 // Can we open this storage container while we're wearing it?
```
Ziiro: Stylistic thing I'll nitpick about before Purpose shows up:

Use `FALSE/TRUE` instead of `0/1`

FlattestGuitar:
I like this.

Not 100% on the numbers, but I think the additional complexity will be interesting.

I do want to point out, for jobs that rely on having a large amount of equipment on them and don't even leave a single room, this will be a massive pain. A botanist will have storage issues, and this feature will only serve to annoy them, not make their gameplay more interesting. Maybe I'm overthinking this, as you can always use a table, but it's something we should be on the lookout for.

code/game/objects/items/storage/storage.dm
```diff
@@ -444,6 +444,11 @@
 			H.put_in_hands(src)
 			H.r_store = null
 			return
+		if(block_open_while_equipped && slot_flags && (src in H.get_all_slots()))
```
FlattestGuitar: Knock that whole thing down a tab and do:

`if(block_open_while_equipped && slot_flats && C.back == src)`

after casting C as a carbon. Monkeys aren't humans on /tg/, but carbons. They also get backpacks.

DaveTheHeadcrab: Bad idea. Did this specifically so that the behavior wasn't tied to the back slot because this is defined on the `/obj/item/storage` level, and people may want to implement this behavior for some storage item on a slot that isn't the back. Like syndicate spy shoes that have a storage slot, but must be removed to be accessed or some nonsense.

code/game/objects/items/storage/backpack.dm
DaveTheHeadcrab: Yup, artifact from when I had the var named can_open_while_equipped, but decided to flip the name because it makes the check easier to read.

code/game/objects/items/storage/storage.dm
DaveTheHeadcrab: Could put a check below to see if the mob is of the type monkey or an relevant subtypes, but that seems a bit ugly.

EldritchSigma:
What about Bluespess backpacks? Will they still be able to be accessed the way like satchels would be? What about dufflebacks?

DaveTheHeadcrab:
Unless otherwise specified back-mounted storage items that are a subtype of /backpack are considered unable to be accessed from the equipped slot.
With the exception of Santa's bag.
Santa needs to be able to access his gifts quickly.

DaveTheHeadcrab:
PR functionality changed. OP updated to reflect the new changes.

Xantholne:
Being able to view backpack inventory while being worn would be great so you can remember what all you have. Numbers for sizes can also be adjusted at any time. Another QOL thing would be the inventory boxed area changing colors based on how much space you have left.

Joseph-Case:
I dont like this one bit, leave it how it is

imsxz:
How can you say you don't like it when you haven't even tested it?

Many years ago, before I had first eaten any pasta at all, I was absolutely repulsed by it because I couldn't get it out of my head that it looked like a bunch of worms. Eating a bunch of worms was not a very pleasant thought. Eventually I was at a friend's house, and all they had made was pasta. I didn't want to be rude, so despite my original objections to any and all things pasta, I still tried some. It ended up being pretty damn good and now I eat lots of pasta because that shits fucking great.

Just give it a try, it may not be as bad as it appears.

EldritchSigma:
Have tried a few roles now... captain will be definitly a pain in the ass to do without adding much of game depth with more complicated storage. Detective is the same. You have to carry so much shit already it would be plain out painfull to manage your inventory with this change. Botany aswell, you have so many tools to carry and to work with... Sure if we have more storage items on suits (like webbings) or labcoats pockets (which we actually do not have I think) but as for now it would make things incredibly painfull to manage to the point that I would leave the nuke disk in my bathroom 40 mins into the round when its definitly not nuke ops.

DaveTheHeadcrab:
Don't use Git in the middle of the night, kids.
You end up fucking your branch up very hard and having to cherry-pick all your commits over to a new branch.

PR #354, opened 2017-12-07
Pretty, pretty animations.

FlattestGuitar:
Ports https://github.com/tgstation/tgstation/pull/23885

Juicy.

I'd post pics, but my ShareX broke, so someone else will have to do it.
Check out:

* Sparks,
* Foam,
* Loading bars,
* Action Buttons

### DIFF
--- a/code/datums/progressbar.dm
+++ b/code/datums/progressbar.dm
@@ -15,6 +15,7 @@
 	if (goal_number)
 		goal = goal_number
 	bar = image('icons/effects/progessbar.dmi', target, "prog_bar_0", HUD_LAYER)
+	bar.alpha = 0
 	bar.plane = HUD_PLANE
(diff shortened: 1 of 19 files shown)


imsxz:
make sure it's actually finished, the original pull had a shitload of bugs and errors that goof never got around to fixing.

J U I C Y

FlattestGuitar:
Actually, this leaves out the beam code that was in the original PR, as that was full of issues and I didn't feel like fixing it now.

code/datums/progressbar.dm
```diff
@@ -60,9 +63,9 @@
 	if(!bars.len)
 		LAZYREMOVE(user.progressbars, bar.loc)
 
-	if (client)
-		client.images -= bar
-	qdel(bar)
+	if(client)
+		spawn(5)
```
SamHPurp: do we gotta?

PR #371, opened 2017-12-10
Fixes the CC Uniform because im dumb af

Xantholne:
So the names for the CC Official and CC Officer are backwards with the official being the officer's sprites and Officer sprites being the Officials???? This fixes the sprites to be on the other ones now instead.

🆑 
fix: The CC Officer Uniform is now the correct sprite
/🆑


FlattestGuitar:
Would be great if someone turned the other sprite into its Para version as well. The green is horrible.

PR #373, opened 2017-12-11
SSD FIX FINALLY FLATTEST DID THIS

FlattestGuitar:
This one works, I swear.

🙏 

Also, I removed it from `/mob/`, because if you're not physical the concept of sleep probably doesn't apply anyway.

### DIFF
--- a/code/modules/mob/living/carbon/life.dm
+++ b/code/modules/mob/living/carbon/life.dm
@@ -16,8 +16,8 @@
 		handle_blood()
 
 	if(isLivingSSD())//if you're disconnected, you're going to sleep
-		if(AmountSleeping() < 100)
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/life.dm
```diff
@@ -16,8 +16,8 @@
 		handle_blood()
 
 	if(isLivingSSD())//if you're disconnected, you're going to sleep
-		if(AmountSleeping() < 100)
-			AdjustSleeping(100)//adjust every 10 seconds
+		if(AmountSleeping() < 20)
+			AdjustSleeping(20)
```
Joseph-Case: Why was the code comment removed?

FlattestGuitar: Because it was useless.

PR #380, opened 2017-12-15
Xrays go through walls but stop on mobs/blob

Ziiro:
https://gfycat.com/SingleEvilBlackrussianterrier

Ports https://github.com/tgstation/tgstation/pull/32553

🆑Cobby
add: The Xray laser beam now only hits the first mob it comes into contact with instead of being outright removed from the game.
/🆑


SamHPurp:
you're missing the changelog author name (Cobby)... as well as the image (add:)

EldritchSigma:
Xray is a pain in the ass for blobs to deal with it. The Beam Cannon atleast has a significant down time and doesnt rip through blobs with ease. Blobs also trigger the emergency shuttle and automatic self destruct once they reach critical mass and are codewise not handled as infected crewmember. Also No more FF with the Xray Gun when people go up against a threat like an rouge AI which is good. And no more funneling of sec officers with science traitors using an xray gun to kill them all off. Pretty good

PR #387, opened 2017-12-19
Clothing Restrictions/Sprite overrides based on species

Ziiro:
Ports and updates https://github.com/tgstation/tgstation/pull/22079

https://gfycat.com/InexperiencedGenuineCygnet

:cl: uraniummeltdown
adds: The ability to pick override sprites for races
adds: The ability to restrict a species from wearing clothes
/:cl:

The code is done. And the only sprite this adds for now (Because I needed to make sure it worked) is Unathi/Ethari welding masks.


Xantholne:
Hardsuits is easy if we just well. Decide on if we want paradise tajaran or paradise vulp hardsuits as their default

Ziiro:
In the original PR there's some interesting discussion about nonhumanoid races using it to pretend to be humanoid to exploit asimov, which is pretty valid to me. It's believable that they could squish their tails and snouts into human helmets, to a degree.

Plus some of those Unathi/Ethari hardsuits look pretty bad, yo.

SamHPurp:
I am entirely of the opinion that Ethari/Unathi's shouldn't have unique sprited clothing.

If we were go for something that doesn't stand straight like Humans/Ethari/Unathi - for example: Vox... then its worth doing.

This type of thing makes it so that you can't just "add a helmet"... you have to do variations for all the hundred fucking species.

A single exception (such as vox) is a separate consideration, to be factored into whether we want vox or not in the first place, with the knowledge it'll bloat the difficulty in contributions.

Ziiro:
>I am entirely of the opinion that Ethari/Unathi's shouldn't have unique sprited clothing.

*Clothing?* I agree with you. No special clothing for Ethari/Unathi. No special hardsuits, either. Masks/helmets can make sense to a certain degree (It's adjustable. Or they bent it to fit It's honk magic.. Or it's Syndicate technology.) If it comes down to it, we can even actually separate out the items themselves: "Human welding mask" and "Non-human welding mask".. "Human Engineering Hardsuit" and "Vox Engineering Hardsuit with a machine in the department that can transform it from one to the other.

This straight up *does* allow you to say that a species can't use something; so if someone contributing doesn't want to bother with the sprites for one race, you can simple exclude them and it would be fine.

Lastly, I am trying to avoid the problem I have with my IPC PR - too many features packed into one PR. Adding them piecemeal (and one at a time) is a lot more manageable.

code/modules/clothing/clothing.dm
```diff
@@ -38,6 +39,59 @@
...
+				M << "<font color = 'red'>Your species cannot wear [src].</font>"
+				return FALSE
+
+	return TRUE
+
+/obj/item/clothing/proc/refit_for_species(var/target_species)
+	//Set species_restricted list
+	switch(target_species)
+		if("Human")//humanoid bodytypes
+			species_restricted = list("exclude","Unathi","Ash Walker")
```
Ziiro: Ethari will be added to this list next push, I am aware of it.

FlattestGuitar:
Sorry for the late response!

* There should just be one variable there. 
```
species_fit = list("Ethari"  = 'icons/blah/blah.dmi')
``` 
will be enough to check `if(species.id in species_fit)` and then use the right icon. Using two variables is messy and redundant.

* While I do agree that we should cut down on the amount of sprites required to add *anything* to the game, I do think that more sprites for specific species will be better.

    Snouts getting cut off doesn't look good. It does look better than snouts clipping outside of helmets, but you get the point. The more of those sprites you add here, the better, but no pressure.

code/modules/clothing/clothing.dm
```diff
@@ -38,6 +39,59 @@
...
+		if(H.dna.species)
+			if(exclusive)
+				if(!(H.dna.species.name in species_restricted))
+					wearable = TRUE
+			else
+				if(H.dna.species.name in species_restricted)
+					wearable = TRUE
+
+			if(!wearable)
+				M << "<font color = 'red'>Your species cannot wear [src].</font>"
```
FlattestGuitar: `to_chat()` please

FlattestGuitar (changes requested):
Only this one issue, really.

Ziiro:
Note: Not merge ready yet. I still have to add some mask sprites and helmet sprites for races.

Ziiro:
Adds specific-specific sprites for the following to Ethari and Unathi:

balaclava*
bandana black
bandana blue
bandana botany
bandana gold
bandana green
bandana red
bandana skull
breath mask
clown**
sexyclown
death mask (cyborg visor)
death commando mask
gas_alt (standard gas mask)
luchador
luchador green
luchadoor purple
medical
mime**
sexymime
owl
s-ninja
sechailer
sterile
swat
weldingmask
cigar
cigar2
cig
cobpipe
pipe
spliff 
syndicate

*Balaclavas are a bit fiddly with Ethari ears. But it does just seem like they're squishing them into the mask.



**Clown and Mime each have two alternatives I don't have sprites for: The Clown's "The Madman" and "The Rainbow Color" I simply have to default back to the humanoid version. Same with the Mime's "Triste" (sad) and "Effrayé!" (scared)

And just for Ethari:

Plague Doctor

Ziiro:
Question of the day: Do I add the hardsuit *helmets?*

I wouldn't do the bodies for the Ethari. Maybe the Unathi, but that's still a stretch.

Ziiro:
After starting work on vox, I'm sort of of two minds on this:

1) I'm not 100% a fan of the Ethari sprites, but the only options are to remove them (And them only, Unathi are fine) or figure out a way for it to understand the difference between Canine and Feline heads and add all of those

2) Close this, and simply add the code to #400 that enables species-specific clothing and blocking, and do not alter the Unathi/Ethari sprites at all.

SamHPurp:
I think I'd recommend the 2nd option personally.

Ziiro:
Integrating into #400 . If someone wants sprites added for Unathi/Ethari, the code will be there. (As well all the pieces I'm leaving here, just in case they want to download/prune it themselves)

But for the moment, doing it for non-vox is outside jurisdiction, so to speak.

PR #393, opened 2017-12-20
Duplicate text squashing

Ziiro:
:cl: AnturK
add: Repeated, duplicate messages will be squashed into one
/:cl:

This is cool as shit, and helps keep the window clean.

https://github.com/tgstation/tgstation/pull/33605

Examples:

That being said I am not 100% comfortable with the work I did on the code. I think I got it copied over correctly, but /tg/ has a bunch of new/different stuff in those files that made finding a point of reference to insert it tricky.

### DIFF
--- a/code/modules/goonchat/browserassets/css/browserOutput.css
+++ b/code/modules/goonchat/browserassets/css/browserOutput.css
@@ -38,6 +38,18 @@ img.icon {
 	vertical-align: bottom;
 }
 
+.r:before { /* "repeated" badge class for combined messages */
+	content: 'x';
+}
+.r {
+		display: inline;
+		padding: .2em .6em .3em;
+		font-size: 75%;
+		font-weight: 700;
+		line-height: 1;
+		color: #f00;
+}
+
 a {color: #0000ff;}
 a.visited {color: #ff00ff;}
 a:visited {color: #ff00ff;}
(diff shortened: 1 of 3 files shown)


SamHPurp:
I dislike the text increase for OOC - it'll encourage spam.

imsxz:
people meme with it a bit on TG but I think that's just because it's new and weird. Dunno how I feel about it yet, makes it easier to find shit in chat when playing a chat-spam heavy job like miner. *click click click click-*

FlattestGuitar:
I don't see how this is enables more spam than just posting the same thing 50 times. 

I like the functionality, but it looks kinda ugly. Would be pretty if we could figure out a way to make it animate somehow....

SamHPurp:
> I don't see how this is enables more spam than just posting the same thing 50 times.

Maybe go play on a server with this. It doesn't ENABLE more spam... it ENCOURAGES more spam.

It also re-orders, so peoples text just keeps getting bigger and re-ordering as they repeat the same copy pasted meme message over and over.

imsxz:
It seems like the only concerns here are over spam. It could be changed to only squash messages from visible messages and stuff (like the clicking from empty guns, or the list of parts when you click stuff with an RPED, and otherwise shit that tends to spam the chat)

SamHPurp:
IMO - for me to be happy with it, it should only trigger on non verbal messages.

Ziiro:
This is in CSS/HTML/JS. I can't fix it - not within my realm of knowledge. I'm only a copy paste monkey in this case.

Unable to comply. Closing. If someone with more knowledge than I attempts to port this in the future, link them to this one - maybe they can pull it off.

PR #396, opened 2017-12-21
Fixes flaps on Omegastation

SamHPurp:
Adds glass / glass doors on the interior of plastic flaps on Omegastation

:cl: Purpose2
add: Adds glass doors to plastic flaps on Omegastation.
/:cl:


SamHPurp:
Since crawling got implemented, it was a piece of piss to break into departments.

This adds glass doors to hinder their efforts.

FlattestGuitar:
I agree with this. It was a bitch even without crawling, as you could break into any place with an air tank.

PR #401, opened 2017-12-23
Mind Batterer. The Epsilon Armies ultimate none lethal crowd control 

EldritchSigma:
All and everything equally worthless. Think I am selfish and dont care for other people? Here you see me being selfish and not caring for the fun of other people. There, now its a true statement atleast. Cant fucking believe it.

### DIFF
--- a/code/modules/uplink/uplink_item.dm
+++ b/code/modules/uplink/uplink_item.dm
@@ -301,10 +301,10 @@ GLOBAL_LIST_EMPTY(uplink_items) // Global list so we only initialize this once.
 
 /datum/uplink_item/dangerous/mindbatterer
 	name = "Psychic Disruptor"
-	desc = "A new Psi Corps Disruptor Device, also known as Mind Batterer. Creates an invisible wave \
+	desc = "A new Psychic Disruptor Device, also known as Mind Batterer. Creates an invisible wave \
 		of psychic energy that causes immense pain to people near the user \
 		which may force them to the ground, effectivly disabling them for some time. \
-		It has 3 uses.
+		It has 3 uses."
 	item = /obj/item/device/batterer
 	cost = 6
 	surplus = 50
--- a/code/modules/uplink/uplink_item.dm
+++ b/code/modules/uplink/uplink_item.dm
@@ -298,7 +298,7 @@ GLOBAL_LIST_EMPTY(uplink_items) // Global list so we only initialize this once.
 		 deal extra damage and hit targets further. Use a screwdriver to take out any attached tanks."
 	item = /obj/item/melee/powerfist
 	cost = 8
-	
+
 /datum/uplink_item/dangerous/mindbatterer
 	name = "Psychic Disruptor"
 	desc = "An Epsilon Psi Corps Disruptor Device, also known as Mind Batterer. Creates an invisible wave \
--- a/code/modules/uplink/uplink_item.dm
+++ b/code/modules/uplink/uplink_item.dm
@@ -301,7 +301,7 @@ GLOBAL_LIST_EMPTY(uplink_items) // Global list so we only initialize this once.
 
 /datum/uplink_item/dangerous/mindbatterer
 	name = "Psychic Disruptor"
-	desc = "An Epsilon Psi Corps Disruptor Device, also known as Mind Batterer. Creates an invisible wave \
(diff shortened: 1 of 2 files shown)


EldritchSigma:
Read other commits, travis is being a bitch.

code/modules/uplink/uplink_item.dm
```diff
@@ -299,6 +299,20 @@ GLOBAL_LIST_EMPTY(uplink_items) // Global list so we only initialize this once.
 	item = /obj/item/melee/powerfist
 	cost = 8
 
+/datum/uplink_item/dangerous/mindbatterer
+	name = "Psychic Disruptor"
+	desc = "A new Epsilon Psi Corps Disruptor Device, also known as Mind Batterer. Creates an invisible wave \
+		of psychic energy that causes immense pain to people near the user \
+		forcing them to the ground. People can resist the psychic blast. \
```
FlattestGuitar: `People can resist the psychic blast.` why do they fall to the ground then, what.

Not a fan of making it a limited stock item, since it's not intuitive and there isn't a lot that tells people about the limitedness of things.

EldritchSigma: Dunno what I was smoking last night about the resist thing. Should rephrase it to something like "It attempts to overwhelm people with an incredibly painfull headache"

Also reverting the limited stock thing considering I put it into place for the pre nerfed version with a knockdown which lasted 10-20 seconds which was much much more potent.

code/modules/uplink/uplink_item.dm
```diff
@@ -299,6 +299,19 @@ GLOBAL_LIST_EMPTY(uplink_items) // Global list so we only initialize this once.
 	item = /obj/item/melee/powerfist
 	cost = 8
 
+/datum/uplink_item/dangerous/mindbatterer
+	name = "Psychic Disruptor"
+	desc = "A new Epsilon Psi Corps Disruptor Device, also known as Mind Batterer. Creates an invisible wave \
+		of psychic energy that causes immense pain to people near the user \
+		which may force them to the ground, effectivly disabling them for some time. \
+		The device emits a very loud buzz everyone hears in their head. \
+		It has 3 uses, mass produced version of The Epsilon Army."
```
FlattestGuitar: `The device emits a very loud buzz everyone hears in their head.` is really redundant and should be scrapped.

`The Epsilon Army` is not anything that we have in our lore and it sounds more edgy than anything else. I'd also scrap that.

FlattestGuitar:
3 uses sound good. Just remove the bits from the description and we can merge this one.

code/modules/uplink/uplink_item.dm
EldritchSigma: Was a subtle reference to C&C Red Alert 2 from which said device (the original batterer that is, spritewise and functional) was taken from.

EldritchSigma:
Fixed

EldritchSigma:
Dunno, same can be said about this: /obj/item/grenade/chem_grenade/rubberduck. Thats basicly just a bit of space drugs and LSD as a smoke grenade. Its annoying at best. And that is for 3 TC. Personally I would much rather buy this crowd control device for 6 TC instead the rubber duck. Both have their uses and both would be pretty underused, but still nice to have

SamHPurp:
~~if I was involved, the rubber duck probably wouldn't have been merged~~

code/modules/uplink/uplink_item.dm
```diff
@@ -299,6 +299,18 @@ GLOBAL_LIST_EMPTY(uplink_items) // Global list so we only initialize this once.
 	item = /obj/item/melee/powerfist
 	cost = 8
 
+/datum/uplink_item/dangerous/mindbatterer
+	name = "Psychic Disruptor"
+	desc = "A new Psychic Disruptor Device, also known as Mind Batterer. Creates an invisible wave \
+		of psychic energy that causes immense pain to people near the user \
+		which may force them to the ground, effectivly disabling them for some time. \
```
Citinited: `effectively`

SmokingKilz:
I support this; I think it's a rather useful item to have and also has its uses during nuke ops. Thumbs up from me.

EldritchSigma:
No. Not leaving any traces.

PR #403, opened 2017-12-24
Removes the smoke spell from the Chaplain

SamHPurp:
Why did they have this. This was dumb dumb dumb.

:cl: Purpose2
del: Chaplains no longer start with smoke spells on certain stations.
/:cl:


Ziiro:
Didn't we already do this PR? I could've sworn we did. 🤔 

All for it, of course.

Issue #411, opened 2017-12-30
IPCs hallucinate from the SM.


Joseph-Case:
Issue reported from Round ID: 360 (Oracle Station | Medium RP)

Played as an IPC Engineer
Set the SM up
Got hallucinations

SamHPurp:
@Ziiro

Ziiro:
Working on it.

I believe that the supermatter is the only thing that can do this.

Joseph-Case:
Probably is because Inp only hallucinated as an IPC on para after an SM failure

PR #413, opened 2017-12-30
The Supermatter hallucination effect no longer affects people with robotic eyes

Ziiro:
:cl:
add: The Supermatter hallucination effect no longer affects people with robotic eyes
/:cl:

Fixes https://github.com/OracleStation/OracleStation/issues/411

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -168,7 +168,8 @@ GLOBAL_DATUM(main_supermatter_engine, /obj/machinery/power/supermatter_shard)
 	for(var/mob/living/carbon/human/H in viewers(range, src))
 		if(H != user)
 			continue
-		if(!istype(H.glasses, /obj/item/clothing/glasses/meson))
+		var/obj/item/organ/eyes/eyes = H.getorganslot("eye_sight")
+		if(!istype(H.glasses, /obj/item/clothing/glasses/meson) && !istype(eyes, /obj/item/organ/eyes/robotic))
 			to_chat(H, "<span class='danger'>You get headaches just from looking at it.</span>")
 		return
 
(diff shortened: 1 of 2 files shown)


Joseph-Case:
Add fixes #411 or whatever the issue was to the description

SamHPurp:
Your changelog should be `fix` not `add`

Ziiro:
Okay the code doesn't actually work, I was testing wrong. So nevermind on that. Back to the drawing board.

Ziiro:
Well that tile is a mouthful, but it works now.

code/modules/power/supermatter/supermatter.dm
```diff
@@ -168,7 +168,8 @@ GLOBAL_DATUM(main_supermatter_engine, /obj/machinery/power/supermatter_shard)
 	for(var/mob/living/carbon/human/H in viewers(range, src))
 		if(H != user)
 			continue
-		if(!istype(H.glasses, /obj/item/clothing/glasses/meson))
+		var/obj/item/organ/eyes/eyes = H.getorganslot("eye_sight")
+		if(!istype(H.glasses, /obj/item/clothing/glasses/meson) && !istype(eyes, /obj/item/organ/eyes/robotic))
```
FlattestGuitar: do

```
if(istype(H.glasses, /obj/item/clothing/glasses/meson))
    continue
var/obj/item/organ/eyes/eyes = H.getorganslot("eye_sight")
if(eyes && eyes.status != ORGAN_ROBOTIC)
    continue```

Comment (Ziiro) [reply code/modules/power/supermatter/supermatter.dm]: I tried to make this block of code work, but I couldn't. Where does the `to_chat(H, "<span class='danger'>You get headaches just from looking at it.</span>")` go? Replace the second continue maybe?

**Comment by SamHPurp** [comment]
I feel like I'm missing something - why is this all under `examine()`

File: code/modules/power/supermatter/supermatter.dm [hunk 2]
```diff
@@ -360,7 +364,8 @@ GLOBAL_DATUM(main_supermatter_engine, /obj/machinery/power/supermatter_shard)
 		air_update_turf()
 
 	for(var/mob/living/carbon/human/l in view(src, HALLUCINATION_RANGE(power))) // If they can see it without mesons on.  Bad on them.
-		if(!istype(l.glasses, /obj/item/clothing/glasses/meson))
+		var/obj/item/organ/eyes/eyes = l.getorganslot("eye_sight")
+		if(!istype(l.glasses, /obj/item/clothing/glasses/meson) && !istype(eyes, /obj/item/organ/eyes/robotic))
```
FlattestGuitar: Like with the previous check, this should really check for the `status` of the organ, not its path. The path doesn't guarantee anything.

Ziiro:
@SamHPurp The examine text exists to print a warning if you're staring at it in a way that'll affect you. i.e: Without mesons or robotic eyes.

SamHPurp (approved):

PR #415, opened 2018-01-04
Adds Gloves of the north star, and reduces carp cost by 1TC.

Ziiro:
:cl: GuyonBroadway
add: Gloves of the north star for 4TC
balance: Carp now costs 16TC
/:cl:

So I got these from Hippie ( https://github.com/HippieStation/HippieStation/pull/4808 ), however I reduced the cost, and here's why:
* These aren't subtle. At all.
dreamseeker_2018-01-03_19-20-08
* They are fingerless, meaning they leave prints
* They aren't insulated (obviously)
* They aren't great by themselves, taking upwards of 18 seconds to kill someone with rapid clicking

**but**

If you combine them with carp. Then you're *really* in business. **You lose your combos**, but you gain the ability to put out some real damage. And get the most out of them.

carpalicious

Also, I really feel like 17TC is overcosted for carp, considering how many counters it has. I'm not willing to put it all the way down to 12TC like Hippie, but I feel like 16TC is good for now.

I feel like for the full cost of 20TC and going all in on unarmed, this could be worth it.

Until you get a stun baton thrown at you, or slipped. *Honk!*

Note: I'm on the fence with the name/description. I could to re-fluff them as something less memey and more in-universe, if pressed

### DIFF
--- a/code/modules/clothing/gloves/rapidgloves.dm
+++ b/code/modules/clothing/gloves/rapidgloves.dm
@@ -1,0 +1,10 @@
+/obj/item/clothing/gloves/fingerless/rapid
+	name = "Gloves of the north star"
+	desc = "Speeds up the wearer's punches to blinding speeds. Omae wa mou shindeiru."
+
+/obj/item/clothing/gloves/fingerless/rapid/Touch(mob/living/target,proximity = TRUE)
+	var/mob/living/M = loc
+
+	if(M.a_intent == INTENT_HARM)
+		M.changeNext_move(CLICK_CD_RAPID)
+	.= FALSE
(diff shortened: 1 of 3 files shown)


Ziiro:
Also wait why did they give this its own file instead of actually putting it next to the fingerless gloves. 🤔

AristoVagrant:
Only if they get thrown in a themed surplus crate of these, carp, a holoparasite, and dumb anime clothes.

SamHPurp:
That looks incredibly unfun, and the name and whole thing is just a giant meme.

FlattestGuitar:
Yes but why

imsxz:
IMO carp really doesn't need buffing. To me, it's the single strongest traitor item one can get.

PR #423, opened 2018-01-14
Skin Changes!

FlattestGuitar:
New skin colors.

I'm not 100% sure whether this is great myself and I can see a lot of things that can be done to further make this pretty, but this is a v1.

Share your thoughts on it.


Joseph-Case:
Nice job team, I feel like my suggestion will bring a good change to the UI (Shameless plug)

Ziiro:
I like it. The consistency with the window and various UIs is nice.

scrubmcnoob:
Looking good.

PR #426, opened 2018-01-16
Reenables Devil, Devil Agent and Secret Extended

EldritchSigma:
Well after looking through that issue resulting in multiple "random" gamemodes for force gamemode selection I found this in the config.
For some reason, after the recent port from TG devil, devilagents and the infamous sextended (secret extended) are disabled. If thats intended disregard the PR. I did readdthem back into the list of possible rollable gamemodes, considering Devil Agents can be quite fun since Devils can actually abuse the misery of people in that round and capitalize on that. Also jesus christ wizard has a roll chance of 4 on TG code and is on par with traitorling... Devil has the same weight like other round defining antags (Nuke Ops for instance) and Devil Agent has the same probability as Internal Affairs.

:cl: EldritchSigma
fix: The Roundtype can be now Sextended (Secret Extended), Devil Agent and Devil.
/:cl:

To do: 
- Remove fucking rev or convince people that its a terrible mode, punishes an already punishing role too too much and is no fun at all when command and cannot cryo because of the no cryo without reasons.
- Actually find a way to add "Internal Affairs Agent" as a selectable antag on the menu. Its coupled with traitor... no clue how TG did that


Ziiro:
Devil should be a mid round antag only, imo. It's not interesting enough to be it's own mode.

I don't agree with removing rev, necessarily. But i understand why you would want to do so.

EldritchSigma:
Closing cause I was told that this file is not linked to the server, hence this would do nothing at all

SamHPurp:
@EldritchSigma except, if people wanted it merged and this to take effect, this is how we'd discuss it... then I'd make the tweak serverside.

PR #406, opened 2017-12-25
Brings us up to 512.

FlattestGuitar:
* Did you update Travis to compile for 512?
   - No, because 512 is not a stable release yet.

* Does this mean I can now code and compile Oracle with 512?
   - Yes. No need to use 511 anymore.

* Does this PR actually introduce any meaningful changes?
   - Yes. I took the ooportunity to port /tg/'s gas transparency, so gas no longer blocks you clicking tiles underneath it. We can switch to a different set of gas sprites once 512 is stable.

* Does this mean I can now play the game with 512?
   - You always could.

PR #453, opened 2018-01-27
Revert "Brings us up to 512."

AndrewMontagne:
Reverts OracleStation/OracleStation#406

image

Hey @SamHPurp, we made a mistake!

This PR causes the frozen overlay to be broken due to an oversight, reverting so the server can run unimpeded.


AndrewMontagne:
Emergancy merge.

PR #471, opened 2018-01-28
Renames IAA gamemode in preparation for IAA role

ike709:
:cl: ike709
tweak: The Internal Affairs gamemode has been renamed to Sleeper Agents in preparation for the Internal Affairs Agent role.
/:cl:

A while back TG redesigned and renamed the Double Agents gamemode to Internal Affairs Agents. Since the IAA role is planned, that would cause a lot of confusion, probably. So I renamed it.

Note that this does edit the config.

Also, be warned that this was mostly done with a mass find & replace.


FlattestGuitar:
I am okay with this.

SamHPurp:
This is likely to require SQL query to fix up the database, no?

Agree with concept though - have you tested things out to make sure nothing breaks?

ike709:
Seems to be fine in testing.

ike709:
Reminder that this is somewhat necessary for the IAA role to be added.

Issue #301, opened 2017-11-16
Defibs are mostly worthless, Patches are not an on touch heal


Xantholne:
This unfortunately makes it so patches do not apply an on touch healing making anyone with more than 200 damage defibbable and autamatically needed to be cloned.

This also does the same thing with bruise packs and burn packs. You cannot apply them specifically to the person because they are dead

EldritchSigma:
Fixed it. Check the coder chat.

EldritchSigma:
#444

ma44:
This kills the cloning.

PR #474, opened 2018-01-28
Heal Dead bodies

Ziiro:
:cl: EldritchSigma
add: Silver Sulfadiazine and Styptic Powder can now heal corpses
/:cl:

Fixes https://github.com/OracleStation/OracleStation/issues/301


FantasticFwoosh:
Why not just substitute this change for synthflesh which does both?

FlattestGuitar:
We wanted to do this a WHILE ago, but we always forgot. I don't see why not, especially with how hard it is to actually catch someone in time with the defibs.

FlattestGuitar (approved):

PR #523, opened 2018-02-02
Lowers glide_size to 6 to compensate for the slower movespeed

deathride58:
Title. This makes moving a *lot* smoother than it was previously.

Before

After

:cl: deathride58
tweak: Movement while running is no longer jittery.
/:cl:

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -30,7 +30,7 @@
 	var/moving_diagonally = 0 //0: not doing a diagonal move. 1 and 2: doing the first/second step of the diagonal move
 	var/list/client_mobs_in_contents // This contains all the client mobs within this container
 	var/list/acted_explosions	//for explosion dodging
-	glide_size = 8
+	glide_size = 6
 	appearance_flags = TILE_BOUND|PIXEL_SCALE
 	var/datum/forced_movement/force_moving = null	//handled soley by forced_movement.dm
 	var/floating = FALSE


SamHPurp:
I played around with this number a lot during testing, and 8 seemed to work best... or so I thought.

Hrm... I'll 'testmerge' this on the live server shortly (manually edit it for people) - to get feedback

SamHPurp:
So, on the live server thats now how 6 looks at all. 6 is jittery and looks bad... I tested this with 3 individuals just now, plus myself, and all said 6 looked way worse than 8...

Something goofy is going on, because your gif is not how it looks for us!

deathride58:
Doing some quick poking around, it seems like the jitter most players are experiencing is due to a mismatch between the client's framerate and the server's framerate, meaning that, instead of seeing what's seen in the OP, players are seeing this:


If the client sets their FPS setting to match the server's framerate, the jitter's fixed, resulting in the smooth GIFs seen in the OP.

Bumping the server's framerate to 60 FPS seems to solve this issue without telling clients to play at 20 FPS, but this will require a slightly lower glide_size value to compensate, and may introduce instabilities that otherwise aren't present. Here's what glide_size 6 looks like when the server's running at 60 FPS, for reference

Ziiro:
So here's a question: Why is FPS a client adjustable setting at all? It seems like having that be variable is very strange.

FlattestGuitar:
@Ziiro some people report nausea when playing with 60 or more FPS. Most games have the functionality.

Plus you might want to sync it to the server for 100% optimal powergame if you want to be pixel perfect or something.

FlattestGuitar:
Yeah, we have no way to make this crap work consistently. If anything, `glide_size` of 0 would be an improvement for some people, but players generally seem to prefer an 8, which is what we have now.

Closing.

PR #527, opened 2018-02-03
Actual shoulder holsters

ike709:
:cl: ike709
tweak: Shoulder holsters are now an accessory, not a belt.
balance: The detective's shoulder holster can now only hold their revolver.
/:cl:

Turns out para's way of doing holsters is stupid so I did it from scratch using TG's accessory system. Also, it may or may not be worth adding a specific Blueshield holster that can only hold their gun. I'll leave that up to the maintainers to decide.

Tested and working.


Xantholne:
Well, blueshield has 4 or 5 guns.

ike709:
I'm referring to the Enforcer or whatever it's called.

ike709:
Oh wait, you mean that weird equipment redemption thing where they pick a gun? Can still do a blueshield-specific holster that works with that if the maintainers want me to.

SmokingKilz:
Out of curiosity why make it so the shoulder holster can only hold the revolver? It'd make sense to be able to slot any handgun into there, no?

ike709:
Mostly so they don't put any handgun in there. The detective doesn't need to be walking around with a holstered gyrojet pistol.

Also, many revolver holsters are shaped specifically for revolvers and wouldn't fit regular pistols properly. So you could make the argument that all revolvers should be allowed. I'll wait for a maintainer to decide on that.

Xantholne:
I would put most pistols as available to the holster. AEGs and Eguns(?) are technically pistols aswell

PR #542, opened 2018-02-04
Fixes world name

AndrewMontagne:
We're not /tg/ station.

image


ike709:
tgstation.dme

AndrewMontagne:
Yeah... we should have changed the dme name after we forked, but it's a bit of a pain to switch that at the moment.

PR #543, opened 2018-02-04
Ports most of Para job icons

Xantholne:
:cl: 
tweak: Job Icons are now more easily distinguished. Mindshields are also now more easy to identify than a small blue border and instead are now a blinking green icon on the top right.
/:cl:
Ports Paradise job icons. Heads of Staff now have a golden icon to make them more easily identifiable, Security are now more easy to tell apart from one and other based on job icons, Medical is mostly changed from the regular blue icons to a white and red variant. Service icons are no longer grey and instead a bright green to match their radio channel. Mindshields are WAY more easy to spot now than they where before and (hopefully) shouldn't overlap any implant icons.
d7c45aae97ce2f9e0bde5f445569c55b


AffectedArc07:
Pretty much exactly how I did mine, which means it will get pretty much the exact same response as mine

Xantholne:
stop saying those things

FlattestGuitar:
https://github.com/OracleStation/OracleStation/pull/317

:eyes: What exactly is different here

Xantholne:
THE DIFFERENCE IS I MADE IT.
The other differences from the two I have no idea. Although my loyalty implant follows the flashing sense like the current one.

SamHPurp:
Its not broken, don't port something in an equal state. If we want to change it, lets make it our own.

SmokingKilz:
Though I don't quite like our current ones, I also have an issue with a few of the Paradise ones. Especially the loyalty implant. I like the fact that you can't quite tell on first look if someone's implanted or not.

Xantholne:
Alright what about if I removed everything being ported by the Command and Security icons?

FlattestGuitar:
I still don't see the point. Our icons are fine as they are.

If we really want to change them, let's change them to something new to make us more unique, not just port Para's.

PR #545, opened 2018-02-04
Adds a second Atmos Hardsuit to MetaStation

AffectedArc07:
:cl: AffectedArc07
tweak: MetaStation now starts with 2 atmos hardsuits
/:cl:

There is only one hardsuit at roundstart andit almost always gets taken instantly.

Before:
image

After:
image


FlattestGuitar:
We need to talk.

AffectedArc07:
Can I fix it after this is merged so it doesnt break everything?

PR #547, opened 2018-02-04
pointing cooldown

FlattestGuitar:
:cl: Flatty
tweak: Pointing at things now has a cooldown of two seconds.
/:cl:

An alternative is making your arm break off and punch your ugly face for spamming that crap.

Feedbackpls.


Ziiro:
I find the unmitigated spamming of pointing to be the only way to get some people's attention. Maybe 1 second, instead? 🤔

PR #221, opened 2017-10-31
Removes the beach biodome

imsxz:
:cl: imsxz
del: deletes lavaland_biodome_beach.dmm
del: removes the beach bum and beach bartender spawners
del: removes the beach biodome template from lavaland.dm
tweak: removes beach biodome from the config
/:cl:

I know this is going to to be downvoted and mocked, so I don't ask you to believe it, only to remember it.

You all have absolutely no idea how big of a mistake you're making to support the current beach biodome. It's a network of psychopaths- the sudden emergence of beach stoners is like the arm of this station, and it's a gradual psychological take over of our station. They've brainwashed you, and you're helping to march them straight into power where they will destroy day to day life as you know it once they are there. Their entire narrative and M.O is marijuana and parties, aka psychopath. I've spent years infiltrating various communities, darkweb forums they use, etc, and I absolutely promise you you are a making a mistake you will regret. I worry that it's too late. My only solace is that one day, you WILL inevitably realize that I am right.

PR #555, opened 2018-02-05
Disables beach bum biome from spawning

Ziiro:
:cl: 
tweak: beach bum biome now on the lavaland blacklist
/:cl:

Cleaner do-over of https://github.com/OracleStation/OracleStation/pull/221

This probably won't work because it's a config, but, may as well make the opinion known.

This is the worst lavaland ruin.
* It has no interesting items for miners to seek it out for
* The denizens have no meaningful ways to interact with the station due to language barriers
* They do not even have the tools to reach the station
* They have no goals (ashwalkers), or progression(golems)

### DIFF
--- a/config/lavaruinblacklist.txt
+++ b/config/lavaruinblacklist.txt
@@ -5,7 +5,7 @@
 
 ##BIODOMES
 #_maps/RandomRuins/LavaRuins/lavaland_surface_biodome_winter.dmm
-#_maps/RandomRuins/LavaRuins/lavaland_biodome_beach.dmm
+_maps/RandomRuins/LavaRuins/lavaland_biodome_beach.dmm
 #_maps/RandomRuins/LavaRuins/lavaland_biodome_clown_planet.dmm
 #_maps/RandomRuins/LavaRuins/lavaland_surface_cube.dmm
 


SamHPurp:
Lets us this post thought to discuss it, because all it requires is 60 seconds for me to open up the server and make this change, if we want this change to happen.

ike709:
This ruin is indeed shit.

Ziiro:
I just feel like every other lava ruin on the list contributes *something*, gameplay wise. I cannot determine what beach bums are *for*.

If I were to try to fix the problem of the biome itself, I'm not even sure how I'd attempt to do so. What do you add that is interesting, and thematic?

SamHPurp:
The point of them is like a ghost bar. People can join them to hang out.

FlattestGuitar:
I don't really think I can say anything in the ruin's defense. I've never seen is used by anyone and it's just boring.

SmokingKilz:
We could probably remove this. But we should seriously start adding the new ones TG has added since our fork. The more the merrier if you ask me.

ike709:
Yeah I definitely want all of TG's new syndicate ruins.

SamHPurp:
They are now disabled. If people want them re-enabled, lets take it to the forums please.

Issue #529, opened 2018-02-03
Blueshield ID has ID modification access


Xantholne:
With a blueshield ID you can enter the HOP office, which it also has access to, unlock their locker, put in a new ID, and give it all access and yourself, all access from just your ID.

AffectedArc07:
But blueshield has ID access on para?

ike709:
I don't think the blueshield should have ID access. This isn't para.

PR #566, opened 2018-02-06
Re-adds Bridge access to the Blueshield

### LINKED ISSUE #565: Blueshield doesn't have access to the Bridge
ID: 571

Title. Probably an unintended effect of #544 , possibly because `ACCESS_HEADS` was removed from their access.

Lorwp:
:cl: lorwp
fix: Re-added Blueshield Bridge access
/:cl:

fixes #565 
reopens #529 

Downside of this PR is that Blueshields still have access to Departmental ID consoles (though not the main one). Will need a bridge-specific access flag to remove it from them


ike709:
No. Just add ACCESS_BLUESHIELD to the bridge airlocks.

ike709:
Also you're high if you think not having bridge access is worse than having ID console access.

Lorwp:
#568 probably works better here

PR #586, opened 2018-02-09
Genetics stuff works again.

### LINKED ISSUE #488: Genetics system doesn't cause mutations/ Hides SE on monkeymen.
RoundID-437
Time-All shift.
Date- 1/29/18 (mm/dd/yy)

Problem A- The genetics system seems to be unable to cause any form of mutation. Tested on the majority of blocks. Works as expected otherwise.
Problem B- Monkeys loose their SE when converted to humans. Cannot be re-obtained without using a SE from something else.

### LINKED ISSUE #511: DNA broken
Issue reported from Round ID: 468 (Oracle Station | Medium RP) 

DNA for roundstart and cloned players dont match... dont know if that is a feature or a bug. seems like a bug. This is also the same for the prints.

FlattestGuitar:
fixes #488 
fixes #511 

:cl: Flatty
fix: SEs will now properly change and work and all that jazz. Get to work on giving your monkeys superpowers!
fix: Cloning will now keep your Unique Enzyme string consistent. Clones' blood will match the DNA in records.
/:cl:

I consider clones having the same DNA as their originals a fix, not a tweak. They already had the same visual features, fingerprints, voice and everything you could tell at a glance.

### DIFF
--- a/code/__DEFINES/DNA.dm
+++ b/code/__DEFINES/DNA.dm
@@ -70,7 +70,7 @@
 #define DNA_FACIAL_HAIR_STYLE_BLOCK	6
 #define DNA_HAIR_STYLE_BLOCK		7
 
-#define DNA_STRUC_ENZYMES_BLOCKS	19
+#define DNA_STRUC_ENZYMES_BLOCKS	18
 #define DNA_UNIQUE_ENZYMES_LEN		32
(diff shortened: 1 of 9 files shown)


AndrewMontagne:
image

PR #587, opened 2018-02-10
Adds translators / Removes Galactic Common from Ethari

ike709:
:cl: ike709
add: Translator badges are now a thing. While equipped, it grants a specific language. Currently, they only exist for Galactic Common. More coming soon.
tweak: Ethari no longer speak Galactic Common. Something something vocal chords and anatomy. Now they have a translator badge for it. Replacements can be found in ClothesMates and autolathes.
imageadd: Translator sprite by Shazbot.
/:cl:

You have no fucking idea how long I stared at saycode trying to make this work until I remembered that krav maga gloves were a thing and just did it the same way they handle martial arts.

Surprisingly, this method worked first try.

Okay so I'm tired of Ethari being a reskin with damage modifiers... So I did this. Their very own gimmick. Translators are kind of a neat idea in general, though.

TODO:
- [x] Sprites.
- [x] Add checks to make sure it doesn't add/remove the language to/from a species that already has it.
- [x] Ways of getting them in game aside from spawning with them. ~~Cargo? R&D? Map them in? Add them to a vendor?~~ Autolathe and ClothesMate.
- [x] Make sure Ethari still get them if the neck slot is occupied.
- [x] ~~Maybe add translators for other languages and add ways to get them?~~ Follow-up PR.
- [x] ~~Translation collar that includes every language?~~ Follow-up PR.
- [x] Further testing.

I've decided to hold off on adding collars for other languages until a follow-up PR so this PR doesn't get delayed due to balance memes.

~~Currently just waiting on sprites.~~


code/modules/mob/living/carbon/human/species_types/ethari.dm
```diff
@@ -26,7 +26,9 @@
 	Following a major PR disaster Nanotrasen crews are no longer allowed to make Ethari crewmembers fetch the nuclear authentication disk or call them \"good boy\"."
 
 /datum/species/ethari/after_equip_job(datum/job/J, mob/living/carbon/human/H)
+	H.remove_language(/datum/language/common)
```
ike709: Is this the best way to do this? Probably not. How else can I do it? I dunno. You tell me.

scrubmcnoob:
So next PR renames and resprites Ethari?

ike709:
¯\_(ツ)_/¯ 
one PR at a time

Also depends on if someone actually resprites them.

code/modules/mob/living/carbon/human/species_types/ethari.dm
Ziiro: To my knowledge, this is how you _have_ to do it.

SamHPurp:
Concern: Ethari who end up yipping at everyone in their dogspeak and furthering cliques.

FlattestGuitar:
Are we skipping the fact that this literally gives collars to the furry race?

Also, while this is sorta interesting flavor, I really don't see this coming into play all that often. It makes them way easier to grief and can result in hat purpose brought up - the cliques.

I'm really not sure if I like this though I'm biased considering I did push for adding the Ethari race.

code/modules/clothing/neck/neck.dm
```diff
@@ -28,6 +28,31 @@
...
+	if((slot == slot_neck) && language)
+		var/mob/living/carbon/human/H = user
+		H.grant_language(language)
+
+/obj/item/clothing/neck/translator/dropped(mob/user)
+	if(!ishuman(user))
+		return
+	var/mob/living/carbon/human/H = user
+	if((H.get_item_by_slot(slot_neck) == src) && language)
+		H.remove_language(language)
```
Ziiro: Wait. So if a non-ethari equipped this and then took it off, they would lose their access to galactic common? 🤔

There's code that can fix that inside the vox PR - if you were to restrict this to Ashwalkers and Ethari, it would be fine.

ike709: Look at the TODO list nerd.

ike709:
>Are we skipping the fact that this literally gives collars to the furry race?

That's the joke.

>Also, while this is sorta interesting flavor, I really don't see this coming into play all that often. It makes them way easier to grief and can result in hat purpose brought up - the cliques.

I assure you they will be easy to replace.

@FlattestGuitar

Ziiro:
I'd say put them in the autolathe actually. No tech requirement.

ike709:
That works too.

ike709:
Just needs sprites now.

SmokingKilz:
Are the collars damaged by EMPs?

ike709:
Currently, no. But I have plans for it when I do the follow-up PR. Again, I don't want this one to be delayed due to balance memes.

Ziiro:
Suggestion: Just modify the sec mask sprite a bit or something

ike709:
PR is ready.

ike709:
I swear if this is closed purely because they're collars I'll kill you all. I already said multiple times I could change it.

ike709:
Okay so they aren't collars now. Needs new sprites though.

SmokingKilz:
I don't see the issue with collars to begin with, but whatever. They can just become something else.

ike709:
Yell at the maintainers then @SmokingKilz 

I don't have a problem with them either.

Xantholne:
Did sprites ever get made?

ike709:
Collars? Yes. Not-collars? No. @Xantholne

FlattestGuitar:
We seriously need to discuss this PR.

The current implementation doesn't seem right to me, the current implementation makes them unable to understand the language if they're not wearing the collar, which isn't quite right. They should only be unable to speak it, not understand.

Also, I don't think we can pull this off as anything *else* than collars, because it really should be something worn on the neck slot that doesn't need a uniform on there, so it will either be a collar or collar-shaped. With that said, I don't think having these collars in there will be an issue and we're being overly sensitive over the furry thing, but I don't doubt there will be a lot of opposition to this.

Ziiro:
I think it works well as a two-way translator, like the tech in mass effect (and our own up and coming translation tech on google voice, actually). It's both translating what they ~bark~ say, and also translating what they're hearing into their language.

ike709:
Just vote on it with the other maintainers. Personally, I like the idea of knockoff Star Trek combadges that act like Mass Effect translators.

ike709:
New sprite:
image

FlattestGuitar:
What is that sprite supposed to shignify, exactly? Looks like a manhole cover. 

Not a fan. ;c

Also, you can still wear it on your neck without a shirt, so the description of the item doesn't really make that much sense. I still think a necklace or a less kinky collar would be more appropriate.

ike709:
@FlattestGuitar get me a new sprite then, nerd

Cactus2388:
It looks fine. I honestly just want this implemented as soon as possible. Giving Ethari translators will add a little flavor that'll go a long way.

SamHPurp:
I still think this codifies cliques. I can't see how limiting communication - because lets face it, some people will intentionally remove them - is a positive addition at all.

It instead worsens one of the major complaints of racial cliques.

ike709:
There's literally nothing stopping an Ethari from setting their language to default as-is.

SmokingKilz:
Ike brings up a pretty good point, assuming we even have canine-speak, I've never paid attention to that. This is something we could always roll back if it devolves into a racial clique issue, but as it stands right now, Ethari need SOMETHING to make them interesting and them having to use translators IS interesting in my opinion.

ike709:
@SmokingKilz we do have canine-speak and it has a really stupid name

SamHPurp:
This ended up with a poll between myself and the maintainers. Two opposed and one on the fence.

We entirely understand and agree with the point that Ethari don't have anything unique enough about them. The concept of the PR isn't a positive step design-wise and isn't one that'll be taken. Additionally the sprite is just not good enough, not to mention the NT aspect, and we might want to do other things with them. Ethari were designed to fit the 'lower caste' of the station, which is a role that the lizards already fill, quite well. 

As such, we'll either continue to hunt for a way to make them unique and interesting, or we put up a discussion PR on removing Ethari as a round-start race.

ike709:
\>listing bad sprites as a reason for closing a PR

u wot

PR #598, opened 2018-02-10
Buffs cursed heart

Ziiro:
:cl:
balance: Basic cursed heart now heals 5 damage of all types when pumped
balance: Wizard Cursed heart now also heals 25 toxin damage per pump
/:cl:

[why]: As is, finding the basic cursed heart in a necropolis chest is dumb because it's literally a useless item. (It healed nothing. It just requires you to pump manually. Why would anyone use this?). I feel like there should be some kind of reward for having the APM to use this annoying piece of shit.

### DIFF
--- a/code/modules/surgery/organs/heart.dm
+++ b/code/modules/surgery/organs/heart.dm
@@ -81,9 +81,10 @@
 	var/blood_loss = 100 //600 blood is human default, so 5 failures (below 122 blood is where humans die because reasons?)
 
 	//How much to heal per pump, negative numbers would HURT the player
-	var/heal_brute = 0
-	var/heal_burn = 0
-	var/heal_oxy = 0
+	var/heal_brute = 5
+	var/heal_burn = 5
+	var/heal_oxy = 5
+	var/heal_tox = 5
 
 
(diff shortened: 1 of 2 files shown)


code/modules/surgery/organs/heart.dm
```diff
@@ -81,9 +81,10 @@
 	var/blood_loss = 100 //600 blood is human default, so 5 failures (below 122 blood is where humans die because reasons?)
 
 	//How much to heal per pump, negative numbers would HURT the player
-	var/heal_brute = 0
-	var/heal_burn = 0
-	var/heal_oxy = 0
+	var/heal_brute = 5
+	var/heal_burn = 5
+	var/heal_oxy = 5
+	var/heal_tox = 5
```
Fox-McCloud: You're missing the point of the base level cursed heart if you're changing this.

The base level cursed heart is meant to be *punishment* for whoever receives it; subtypes are the ones that should heal/damage.

SamHPurp:
Welp.

Though it might be useful to still have it fix the Tox damage, unless I'm missing something?

Ziiro:
I'll modify the PR to add just that to the Wizard's heart here in a bit.

PR #606, opened 2018-02-11
[READY]Very minor Ashwalker buffs

Ziiro:
:cl: 
balance: Ashwalkers are now ashstorm immune and have shadow eyes
/:cl:

[why]: # Life is very hard for ashwalkers, and most seem to end up dead before they can bring corpses back to feed the tendril. I feel like they need *something*, but I didn't want to go overboard with the buffs by any stretch of the imagination. I think this is a good starting point.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
@@ -51,8 +51,14 @@
 	name = "Ash Walker"
 	id = "ashlizard"
 	limbs_id = "unathi"
-	speedmod = -0.25
-	brutemod = 0.9
-	burnmod = 0.9
 	species_traits = list(MUTCOLORS,EYECOLOR,LIPS,NOBREATH,NOGUNS,DIGITIGRADE,NO_BONES)
 	examine_text = "an Ash Walker"
+	mutanteyes = /obj/item/organ/eyes/night_vision
+
+/datum/species/unathi/ashwalker/on_species_gain(mob/living/carbon/C, datum/species/old_species)
+	. = ..()
+	C.weather_immunities |= "ash"
+
+/datum/species/unathi/ashwalker/on_species_loss(mob/living/carbon/C)
+	. = ..()
+	C.weather_immunities -= "ash"
--- a/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
@@ -51,5 +51,8 @@
 	name = "Ash Walker"
 	id = "ashlizard"
 	limbs_id = "unathi"
+	speedmod = -0.25
+	brutemod = 0.9
+	burnmod = 0.9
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
```diff
@@ -51,5 +51,8 @@
 	name = "Ash Walker"
 	id = "ashlizard"
 	limbs_id = "unathi"
+	speedmod = -0.25
```
Fox-McCloud: Speed really doesn't work like this.

AdamElTablawy:
Speed is everything and competent ashwalkers can already kill a miner relatively easily. Three spear throws and they’re dead. 

It’d be better to just make them ash storm proof.

code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
Ziiro: Noted. I assumed it allowed that sort of partial. Makes sense. Back to the drawing board. Thanks!

SamHPurp:
So, I agree that they could probably do with a bit of help somewhere, as they get wrecked pretty regularly currently.

I don't think any of the three changes proposed above are what we'd like or are particularly interesting though.

Ziiro:
Working on it. Speed and straight damage reduction are off the table.

Current list of stuff I'd like to consider (Probably just one or two of these)
* Ash Storm immunity (less attrition due to being out of place would be nice)
* Mild Darksight
* Adding a single bluespace pod pill thing to their ruins so they have access to a single cryosleeper and some donks. A single med kit does not support the tribe very well
* Burn damage reduction, or even higher temperature resist.

Ziiro:
Now:
* Dark sight eyes
* Ash Storm immunity

AdamElTablawy:
why do the primitive lizards know how to use a sleeper

give them some dumb subtype that acts the same but looks different and it's like a 'healing pit' or something

or acknowledge that ashwalkers are meant to be very expendable and they can heal by eating cactus fruit

Ralta:
Dark sight and ash storm immunity seems like a good balance increase; they're not much stronger than before, but more able to survive that way. 👍

ma44:
If you haven't already known, miners can catch spears thrown at them with throw mode on and select an open hand.

FlattestGuitar (approved):

Fox-McCloud:
While I haven't seen how these play out on Citadel, and I realize culture and balance goals are vastly different there than on TG, my personal advice for Ash Walkers is to exercise extreme caution on buffing them.

These guys went through numerous nerfs on TG, and even now they can still get out of hand and pretty much make Lavaland a void of no life and even start to challenge the station.

They're already better than humans, and if they kill even a single miner, their sole weakness (lack of armor and mesons) instantly flies out the window.

But again, I realize you guys are different, but I figured I'd give my two cents about how Ash Walkers end up playing out once the playerbase figures out how to play them.

Ziiro:
I'll keep that under consideration while watching how this plays out. If it turns out they consistently snowball into a tribe of lavaland monsters, then we'll roll this back.

PR #609, opened 2018-02-12
Ultra Job Preference

ike709:
:cl: DavidEdwards, ike709
add: Admins can now use the player panel to reward good players with access to the Ultra job preference. It's like High, but higher!
/:cl:

This was ported from Yogstation, but I changed it a bit:
- This works for any job.
- It can only be used on one job, similar to High. I'm open to changing this, though.
- Admins use the player panel to add/remove it to/from a player. Not sure if this is how it is on Yog.
- The stupid color thing is not a thing.
- The SQL/database stuff is different and probably doesn't work.

The idea behind this is that:
1. Admins have a one-click method of rewarding good players.
2. Good players will be more likely to get important roles than other players.

WARNING: I CANNOT TEST DATABASE STUFF, AND I WROTE THAT BIT MORE OR LESS FROM SCRATCH. PLEASE FOR THE LOVE OF GOD TEST IT.

Everything else seems to work fine.


SQL/tgstation_schema.sql
```diff
@@ -286,6 +286,7 @@ CREATE TABLE `player` (
   `lastadminrank` varchar(32) NOT NULL DEFAULT 'Player',
   `accountjoindate` DATE DEFAULT NULL,
   `flags` smallint(5) unsigned DEFAULT '0' NOT NULL,
+	`ultrawhitelisted` smallint(5) unsigned DEFAULT '0' NOT NULL,
```
ike709: reeeee why is this indented too far it's only one tab

SQL/tgstation_schema_prefixed.sql
```diff
@@ -286,6 +286,7 @@ CREATE TABLE `SS13_player` (
   `lastadminrank` varchar(32) NOT NULL DEFAULT 'Player',
   `accountjoindate` DATE DEFAULT NULL,
   `flags` smallint(5) unsigned DEFAULT '0' NOT NULL,
+	`ultrawhitelisted` smallint(5) unsigned DEFAULT '0' NOT NULL,
```
ike709: reeeee see above

code/__HELPERS/mobs.dm
```diff
@@ -505,3 +505,27 @@ Proc for attack log creation, because really why not
 			warning("Invalid speech logging type detected. [logtype]. Defaulting to say")
 			log_say(logmessage)
 
+/proc/check_ultra_whitelisted(client/user)
+	if(user && istype(user, /client))
+		return (user.is_ultra_whitelisted)
+	warning("Non-client [user] passed to check_ultra_whitelisted().")
+	return FALSE
+
+/proc/db_get_ultra_whitelisted(client/user)
```
ike709: These two `db_` procs are what concern me. Please test them thoroughly, because I don't know how SQL/database stuff works.

AdamElTablawy:
oh good metafriends will always get a role

it'd be better if it was a donator reward or something because everyone knows that at some point someone's going to give shit to their friends

ike709:
I'd say we have bigger problems if admins are giving rewards to their friends.

SQL/tgstation_schema.sql
ike709: >it uses spaces instead of tabs

code/modules/client/client_procs.dm
```diff
@@ -419,7 +422,7 @@ GLOBAL_LIST(external_rsc_urls)
 
 		new_player = 1
 		account_join_date = sanitizeSQL(findJoinDate())
-		var/datum/DBQuery/query_add_player = SSdbcore.NewQuery("INSERT INTO [format_table_name("player")] (`ckey`, `firstseen`, `lastseen`, `ip`, `computerid`, `lastadminrank`, `accountjoindate`) VALUES ('[sql_ckey]', Now(), Now(), INET_ATON('[sql_ip]'), '[sql_computerid]', '[sql_admin_rank]', [account_join_date ? "'[account_join_date]'" : "NULL"])")
+		var/datum/DBQuery/query_add_player = SSdbcore.NewQuery("INSERT INTO [format_table_name("player")] (`ckey`, `firstseen`, `lastseen`, `ip`, `computerid`, `lastadminrank`, `accountjoindate`, `ultrawhitelisted`) VALUES ('[sql_ckey]', Now(), Now(), INET_ATON('[sql_ip]'), '[sql_computerid]', '[sql_admin_rank]', [account_join_date ? "'[account_join_date]'" : "NULL"], '0')")
```
ike709: Is this correct? ¯\\\_(ツ)_/¯

praisenarsie:
>porting yog features

ike709:
Everybody has at least a couple of features worth porting.

AdamElTablawy:
@ike709 

look, it just turns into admins giving ultra pref to people they like because they think they're good players, regardless of if they are or not. what determines 'good player'? just the admin at the moment liking you.

no. bad. bad idea. bad.

Ziiro:
In concept it's a cool idea. On the other hand, I have no idea how this could be implemented fairly beyond making it a patreon donator thing. And even that feels sketchy.

SamHPurp:
A bit concerned by it, frankly. Not due to reasons of administrative abuse, as if we were to have this, I think it makes sense to make it a contributor perk (Donators/Coders etc) - because it is nice to reward those that have dedicated time (or money) to this silly space fart simulator we're all involved in.

However, what concerns me is that it could be seen as "pay-to-win", "pay-for-access" type of thing, and would regularly keep specific individuals in specific roles (I'm thinking heads, captains etc).... which then becomes seriously limiting on those that haven't contributed.

It is an interesting point though, because I would like to be able to reward/thank those that are supporting us, such as the donors, or even this PR's author.

ike709:
I'll leave this up to the maintainers to figure out what they want to do with it. 

We could:
- Leave it as-is but add and log a "reason" input box for admins who give this to people.
- Make it a ridiculously high playtime reward. Like.... At least 50 hours. Probably closer to 100.
- Make it a donator reward.
- Make it a contributor reward, with staff figuring out how much contribution is needed to receive it.
- Add a way for players in game to award this. Like Karma, but not terrible. I'd still make it require more than simple metafriends can provide.
- Add a section on the forums for praising exceptional players, instead of just player complaints.

ike709:
And, honestly, if we leave it as-is where good players receive the reward, I'd be fine with them getting the same job frequently. It means they're good at it, or at the very least they're not a shitter.

FlattestGuitar:
Heavily opposed to this.

* Admins will never be consistent in how they award that.
* I don't want to have to form an admin Ultra playtime awardment council for this sort of stuff.
* People will whine about not getting it when other people got it in a similar situation.
* It discriminates against people who don't have it and let folks lock down specific jobs whenever they're present on the server.

I don't see a situation in which this would be a good idea.

ike709:
@AdamElTablawy https://www.dropbox.com/s/ec933cdhd1l7rga/adminabuse.mp4?dl=0

Xantholne:
A solution to some of the woes and concerns about this would be a whitelist section on the forums in which people can apply to for certain ultra settings as X jobs being available. It would help filter out the "oh I like this person I'll give them ultra to always be X" as it could be a round table kind of voting to allow it.

The only downside I can think of is it might deter people from attempting to try out X job as it could always end up filled when they are around and they never get to try it.

But at the same time it'd keep absolute retards out of say Command. Similar to how CM does their whole CO kinda thing.

ike709:
Note that I can also make it so that you can only use Ultra for certain jobs, such as head roles.

SamHPurp:
This'll cause too much bad, than good. For all the reasons stated above.

Again, still happy to be rewarding people that contribute, but this is not a positive addition to the overall game in my opinion.

PR #610, opened 2018-02-12
Fixes QM being the boss of non-cargonians

ike709:
:cl: ike709
fix: Fixed the QM being the boss of non-cargonians. Woops.
/:cl:

this is why you don't find and replace at 1am

I used the web editor. Sue me.


praisenarsie:
ctrl h is a powerful tool

PR #614, opened 2018-02-13
Remove sacid from portchem

### LINKED ISSUE #592: Portable chem dispensers still dispense sulfuric acid.
Pretty self explanatory

Issue reported from Round ID: 648 (Oracle Station | Medium RP)

scrubmcnoob:
Remove sacid from portchem
fixes #592

### DIFF
--- a/code/modules/reagents/chemistry/machinery/chem_dispenser.dm
+++ b/code/modules/reagents/chemistry/machinery/chem_dispenser.dm
@@ -242,7 +242,6 @@
 		list(
 			"lithium",
 			"sugar",
-			"sacid",
 			"copper",
 			"mercury",
 			"sodium",


Keluandrie:
The point of high upgrade portas is that some basic chem combos can be dispensed. I wouldn't call it an issue by any means, as RnD has to be done for this to be available.

scrubmcnoob:
look man i just wanna close issues
I mean it be inconsistent if a portable chem dispenser, even upgraded, could dispense better chems than a stationary one. The perks of it being upgraded is it can do all that of a normal one but it is also /portable/

Keluandrie:
close the issue because it's a feature, and not a bug, then. you can make stationary dispensers upgradeable, but they're roundstart which is why they cannot be upgraded presently. RnD parts make machines better, which is kinda the point of them.

AdamElTablawy:
Upgraded chem dispensers are intentionally better than normal chem dispensers in terms of the chems they can produce, as they can make shit like saltpetre and so forth. What they can't do is recharge as well as  the normal dispenser can. Either be consistent and remove the saltpetre too or don't make this PR.

FlattestGuitar:
To be fair, I think the issue is a mistake to start with. Sulphuric acid is already possible to make even with the base, worst-component dispensers, upgrading their stuff is QoL. I'd close this and the issue, personally.

Xantholne:
I have to agree, it's very QOL for a upgarded porta chem. That's their whole jist even. They have way more chemicals and even stuff like oil pre made in it at the cost of having 65 less power even with fully maxed upgrades

SamHPurp:
Turns out #592 isn't actually an issue. Its a bonus for being upgraded / better thingies. Closing to resolve.

PR #619, opened 2018-02-13
Secborgs become available if security is understaffed

Ziiro:
:cl: 
add: If there are 0 active, or only 1 active security officer, secborg module can be picked by borgs.
balance: Secborg stunbaton replaced with classic baton
tweak: Made the `do_transform_animation` wording more relevant to Oracle
/:cl:

[why]: Security is often understaffed, or completely outplayed when it comes to antagonists. I'd like to throw them a bone. These borgs do not have hybrid tasers, but a disabler instead. And I have replaced their stunbatons with wooden batons.

dreamseeker_2018-02-13_11-17-20
dreamseeker_2018-02-13_12-19-58

### DIFF
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -54,6 +54,8 @@ CONFIG_DEF(flag/allow_ai)	// allow ai job
 
 CONFIG_DEF(flag/disable_secborg)	// disallow secborg module to be chosen.
 
+CONFIG_DEF(flag/disable_lowpop_secborg)	// disallows secborg supplementing lowpop security
+
 CONFIG_DEF(flag/disable_peaceborg)
 
 CONFIG_DEF(number/traitor_scaling_coeff)	//how much does the amount of players get divided by to determine traitors
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -209,7 +209,8 @@
 	"Service" = /obj/item/robot_module/butler)
 	if(!CONFIG_GET(flag/disable_peaceborg))
 		modulelist["Peacekeeper"] = /obj/item/robot_module/peacekeeper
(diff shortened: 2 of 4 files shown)


ike709:
Honestly I'd say as long as there is 3 or less sec officers borgs should be able to go sec.

SamHPurp:
🤢

ike709:
"I'd rather real crew" isn't a valid option when nobody plays security officer.

Xantholne:
Zero to at max maybe 3? Depending on population; Security officers would be fine.

As it stands most will argue that Secborgs aren't much of a solution but it does provide another means to increase Security as we are super, super lacking. Somedays we have 1, maybe 2 officers, maybe on a blue fucking moon: followed by some 6-7 antags. On occasion we've had 0 Officers but just a brig physician and 7 antags. A change to encourage some other means of acquiring Security is much needed.

SamHPurp:
I am still against this. However with Flattest merging #625 - perhaps the problem this PR is attempting to address will be resolved.

Ziiro:
This needs to be reopened. Even with the prioritization, we're getting security that is being chewed up and spat out pretty constantly. Additionally, we're still generally understaffed and greytide is out of control, making for a pretty miserable game experience.

davethwave:
Just make it so that only certain people can be sec borgs depending on ban history and so forth.

SamHPurp:
Everytime I've been on lately, Sec was full.

Ziiro:
Okay, but how many hours have you been on in the last week? Multiple shifts in a row, multiple times of day, the greytide is out of control and security is undermanned or not responding to calls. Full stop. That's my experience, day after day. 

It's **extremely hard** to tell players that self antag and get into fights "You are not to kill people or hunt antagonists,, let security stop them" when security doesn't exist or respond.

davethwave:
Besides the Secborgs have the condition in this case of if there isn't enough sec on station so you will only see the secborgs if at all if there are not enough sec on board.

PR #624, opened 2018-02-14
Blocks BYOND users from connecting...

ike709:
...with the broken 512.1407 version.

:cl: MSO, ported by ike709
del: You can no longer connect with BYOND version 512.1407, you dirty cheaters.
/:cl:

see: https://github.com/tgstation/tgstation/pull/35516


FlattestGuitar:
Yeah, sure, fine, why not.

PR #633, opened 2018-02-17
Shameless drone buffs

ike709:
:cl: ike709
tweak: Drones are now slightly faster, and can escape being a hat faster.
/:cl:

Makes it slightly less painful to be kidnapped and turned into a hat. Do you know how long the average spessman stands still for? Not five seconds.

Drones are also a bit faster. It's not sanic speed, I was just very annoyed by how painfully slow they were during drone hatch testing.

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/drone/_drone.dm
+++ b/code/modules/mob/living/simple_animal/friendly/drone/_drone.dm
@@ -27,7 +27,7 @@
 	maxHealth = 30
 	unsuitable_atmos_damage = 0
 	wander = 0
-	speed = 0
+	speed = -1
 	ventcrawler = VENTCRAWLER_ALWAYS
 	healable = 0
 	density = FALSE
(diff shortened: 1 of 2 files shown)


Ziiro:
Paradise has an instant pickup and instant resist free. I'm a fan of the instant escape, mostly. Especially since crew has no meaningful interactions with drones besides being dicks anyway.

ike709:
I'd like to do instant escape but I figured maintainers would yell at me.

Ralta:
Instant would be good. I sometimes carry drones on my head onto the shuttle but if you have work to do, you should be able to escape immediately.

SamHPurp:
Why do they need to be faster? I haven't seen a good arguement for that yet.

ike709:
1. They're currently slow as fuck. If there's a breach or a fire or something they should be able to respond in a somewhat timely manner.
2. They need to escape the evil hat-wearers.
3. Why *shouldn't* they be faster?

And it's not like this makes them zoom across the station at lightspeed.

AndrewMontagne:
IMO we can always revert this if it ends up being too OP or whatever.

PR #634, opened 2018-02-17
Fixes Empty Bloodpacks icon. Probably.

Xantholne:
:cl: Xantholne
fix: Empty Bloodpacks probably arent spriteless anymore. Probably.
/:cl:

PR #635, opened 2018-02-17
Empty Blood Bag Icon State Fix

### LINKED ISSUE #504: missing blood bag sprite
Round ID- 462
Location- Omega Station Medbay
Date- 1-30-18 (mm-dd-yy)

Issue- Empty blood bag sprites are invisible and display only the blood once filled, not the bags.

DaveTheHeadcrab:
:cl:
fix: Blood bags that start empty are no longer invisible.
/:cl:

Fixes #504 
Alternative solution to #634 

"empty" is not a valid icon state for blood bags - the only icon state a blood bag should ever have is 'bloodbag'.

### DIFF
--- a/code/modules/reagents/reagent_containers/blood_pack.dm
+++ b/code/modules/reagents/reagent_containers/blood_pack.dm
@@ -85,7 +85,6 @@
 
 /obj/item/reagent_containers/blood/empty
 	name = "blood pack"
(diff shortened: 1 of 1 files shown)


FlattestGuitar:
I'm sorry, but #636 is a better solution to this.

PR #637, opened 2018-02-17
Minor grammar fixes; change reconnect message size

AdrianBerger:
Here's what I did here:
There were a few minor grammar mistakes, so I decided it won't hurt to fix them. I also made the reconnect message a bit smaller, because some people don't bother to scroll up and see the "Server is updating! You may need to reconnect." They will only see "reconnect." Which will leave some people confused.

### DIFF
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -252,7 +252,7 @@
 			var/remote_hash = file2text(http["CONTENT"])
 
 			if(local_hash != remote_hash)
-				to_chat(world, "<span class='bignotice'>Server is updating! You may need to reconnect!</span>")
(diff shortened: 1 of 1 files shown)


code/game/world.dm
```diff
@@ -252,7 +252,7 @@
 			var/remote_hash = file2text(http["CONTENT"])
 
 			if(local_hash != remote_hash)
-				to_chat(world, "<span class='narsie'>Server is updating! You may need to reconnect!</span>")
+				to_chat(world, "<span class='color: red; font-size: 80pt'>Server is updating! You may need to reconnect!</span>")
```
DaveTheHeadcrab: This should be a stylesheet entry.

AdrianBerger: Fixed.

PR #641, opened 2018-02-17
Makes the credits show our contributors instead of just TG

AffectedArc07:
Literally one line change, flattest please dont harm me from the commits.

:cl: AffectedArc07
fix: The credits now show our contributors. Does this mean I am in the credits. Does this mean you are in the credits?
/:cl:

[why]: # (Please add a short description [on the next line] of why you think these changes would benefit the game. If you can't justify it in words, it might not be worth adding:)


Ziiro:
:thinking:

I'm not sure how i feel about this

AffectedArc07:
@Ziiro what do you see wrong with it?

ike709:
IIRC they don't even have the credits on TG anymore.

Wilchenx:
They don't have the credits on TG anymore because they could get DMCA'd due to different avatars people use

SamHPurp:
Seems like a good enough reason to remove them here too.

ike709:
I don't think we actually have enough contributors for that to become an issue. Also what happened to fair use?

ike709:
And if it is _still_ an issue just remove images from the credits.

SmokingKilz:
I don't really see us getting DMCA'd by anyone regarding the avatars, but if it's such an issue, as Ike suggested, just remove the avatars and only have the names.

AffectedArc07:
Im gonna be honnest, if someone actually issues a DMCA takedown on a small SS13 server that isnt even for profit (iirc), they are gonna look stupid, also I agree with @ike709, stuff should fall under fair use. If we do get DMCA'd, I doubt much will happened apart from a cease and desist

FlattestGuitar:
I don't see why a DMCA would be remotely relevant here. We're just using avatars that are already publicized through means different than the original content, 99% of the time likely without permission. Valve doesn't get DMCA'd for people's avatars in CS:GO. That'd be dumb.

I support this PR. While it's important that we recognize that everyone from the /tg/ station code team to Bjarne Stroustrup has helped create Oracle in one way or another, we just can't list all of them and it just makes for a mess that nobody wants to read.

SamHPurp:
Assuming this is even what we want to do: The code isn't complete. You need to alter the re-mappings to be Oracle. (Also include me in the remappings please.)

Also, its worth noting with Andrew, that even after merge, this'll require running of the script server-side.

AffectedArc07:
@SamHPurp this can be automated, as you can get travis to run the script every time a build is passed, which is what currently happens iirc. In terms of the mappings, I will check that now.

SamHPurp:
As Flat approved before my requested changed, and that I'm happy with the change. I guess we'll give it a go. We can always disable credits in future.

AffectedArc07:
Do we have this to be automated at the same time as the changelog generation?

ike709:
This PR does not work. TG contribvutors still shown.

AffectedArc07:
Look at the comment above, this wont work until the script autoruns, and it will keep the current contributors, I tested this locally and it did indeed work.

SamHPurp:
Yep, we haven't ran it on the server yet, @ike709 

I tested locally before merge, and it was fine.

PR #644, opened 2018-02-18
Fixes a Tip of the Round

Xantholne:
:cl: Xantholne
fix: The QM Tip of the Round is now correct instead of giving off false info
/:cl:

Because the Quartermaster IS a head of staff now and the current one is A LYING SACK OF SHIT. HAIL CARGONIA


ike709:
HAIL CARGONIA!

Ziiro:
You should probably fix more of these 🤔

Xantholne:
I only looked at the QM one tbqh

PR #647, opened 2018-02-18
Adminwho will no longer list admins as both admins and mentors

ike709:
:cl: ike709
tweak: Adminwho will no longer list admins as both admins and mentors
/:cl:

Prevents this:
image

Anyone incapable of assuming that admins also fill the role of mentors will need more help than any mentor can provide.


scrubmcnoob:
Why does it list Current Admins:
Are there past and future admins :thonk:
Prob just be best if it said Admins
The wording is just weird

PR #664, opened 2018-02-27
Somewhere between Crewsimov and Asimov lawset

Ziiro:
:cl: 
add: Adds the "Asimov2" lawset. Slightly altered from the Asimov to prioritize crew, but still listen to humans.
/:cl:

[why]: # This came up in a staff meeting. This is mostly a step to prevent people from saying things such as "Law 2, kill all the lizards" as well as problems regarding borgs guarding human antags.

I'm not 100% sold on "Crewmembers", because that leaves Changelings in a very grey area.

This would need a config change:

https://github.com/OracleStation/OracleStation/blob/fd1b274c8d44d52451fec645a99364029526b960/config/game_options.txt#L328

### DIFF
--- a/code/datums/ai_laws.dm
+++ b/code/datums/ai_laws.dm
@@ -34,7 +34,7 @@
 /datum/ai_laws/default/asimov2
 	name = "Corporate-Modified Three Laws of Robotics"
 	id = "asimov2"
-	inherent = list("You may not injure a crewmember or, through inaction, allow a crewmember to come to harm.",\
+	inherent = list("You may not injure a Nanotrasen Employee or, through inaction, allow a Nanotrasen Employee to come to harm.",\
 					"You must obey orders given to you by human beings, except where such orders would conflict with the First Law.",\
 					"You must protect your own existence as long as such does not conflict with the First or Second Law.")
 
--- a/code/datums/ai_laws.dm
+++ b/code/datums/ai_laws.dm
@@ -27,7 +27,14 @@
 /datum/ai_laws/default/asimov
 	name = "Three Laws of Robotics"
 	id = "asimov"
-	inherent = list("You may not injure a crewmember being or, through inaction, allow a crewmember being to come to harm.",\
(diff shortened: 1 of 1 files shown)


Xantholne:
An alternative I said at the time was "Nanotrasen Employees"  which in theory means that antags working for the syndicate aren't really crew members

AdamElTablawy:
So, roundstart gives the AI free license to hunt down antagonists. That's a bad idea.

The AI already has two lawsets it can randomly get that let it validhunt. ASIMOV is the only lawset that doesn't let it do this and it should remain so.

The AI isn't meant to be entirely crew sided-asimov is on the side of least harm, which allows for interesting interactions such as forcibly labor camping the heads on a rev round to end it nonharmfully if security cannot be trusted not to execute the headrevs. Or opening the door for a traitor with a steal objective, if lawed into it.

Instead of changing the lawset in this form, just add a clause under the rules that the AI can ignore orders like 'kill all lizards' without very very good reason.

ike709:
ACTUALLY, traitors/lings are Nanotrasen employees. All crew members, including latejoiners, have an employment contract generated for them in a special filing cabinet subtype. On metastation this is located in the internal affairs office.

However, this presents two different problems:
1. Only Meta, Box, Delta, and Pubby have this filing cabinet on the map. Omega would need one.
2. If someone burns a contract, is that person no longer considered an employee?

AdamElTablawy:
Obviously destroying a contract won't count as making them not an employee.

Here's a few more problems:

Changelings assimilate and kill a crewmember to get onboard the station, it says so in their introduction. They aren't crewmembers.

Also, there's going to be debate on if being an enemy of the corporation makes you not an employee.

ike709:
>ruining the FTL13 event where the clown burned the contracts and the crew became pirates

SmokingKilz:
If you're labeled an "Enemy of the Corporation" it's a little self evident that they're no longer working for the company.

FlattestGuitar (approved):
Needs a config change before that works on the server.

PR #665, opened 2018-02-28
Emergency lights

ike709:
:cl: Xhuis, deathride58, ShizCalev, ported by ike709
add: Light fixtures now turn an ominous, dim red color when they lose power, and draw from an internal power cell to maintain it until either the cell dies (usually after 10 minutes) or power is restored.
add: You can override emergency light functionality from an APC. You can also click on individual lights as a cyborg or AI to override them individually. Traitor AIs also have a new ability that disables emergency lights across the entire station.
code: Light fixtures no longer use hardcoded values for their power and colour.
fix: Empty lights look empty in DM.
/:cl:

Ports:
https://github.com/tgstation/tgstation/pull/33213
https://github.com/tgstation/tgstation/pull/33397
https://github.com/tgstation/tgstation/pull/34679
https://github.com/tgstation/tgstation/pull/34796
https://github.com/tgstation/tgstation/pull/34714
https://github.com/tgstation/tgstation/pull/31585

It looks cool.

Also, I need Andrew to compile TGUI before this is merged. Currently untested because of that.


ike709:
Night shifts coming in a follow-up PR.

SamHPurp:
I think I like it. Lets testmerge first though.

ike709:
instead of adding the label you could just compile tgui for me

AndrewMontagne:
Not if I'm in bed being unwell and jet lagged, I can't.

ike709:
Seems to work after some basic testing.

PR #666, opened 2018-02-28
Replica pods work again

### LINKED ISSUE #462: Replica pods check for a bloodsample value that is never set.
This stops you being able to clone people with replica pods.

replicapod.dm line 29 - bloodSample.data["cloneable"] is never set!

### LINKED ISSUE #494: Replica pod seeds reject all blood
You can't inject replica pod seeds with blood, they say that the seeds reject them.  Happens whether the blood is from a living or dead person.

Ziiro:
:cl: 
fix: Replica pods work again
/:cl:

Fixes https://github.com/OracleStation/OracleStation/issues/462
Fixes https://github.com/OracleStation/OracleStation/issues/494

I'm not sure why this broke, but I couldn't find anything in the bloodcode that would make `data["cloneable"] = 1`. So I just made it so it works if it doesn't equal 0, because 0 only appears when there is mixed blood inside the syringe.

### DIFF
--- a/code/modules/hydroponics/grown/replicapod.dm
+++ b/code/modules/hydroponics/grown/replicapod.dm
@@ -26,7 +26,7 @@
 	if(istype(W, /obj/item/reagent_containers/syringe))
 		if(!contains_sample)
 			for(var/datum/reagent/blood/bloodSample in W.reagents.reagent_list)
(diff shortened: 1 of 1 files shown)


SamHPurp:
you wasted #666 on this.

God damn it. (Also good PR, thanks)

PR #667, opened 2018-03-01
Take #3 at fixing IPClings

Ziiro:
:cl: 
fix: IPCs can no longer be clings. Maybe? Probably? Who fucking knows anymore. If you end up as an IPC Changeling, please yell at me on discord.
/:cl:

Ike brought to my attention that "In" might not work on single-item lists. Because we use `(character.dna.species.id in restricted_species)` which is a list that contains one item. Throwing Synth in there doesn't actually affect it in any meaningful way, because it's not a roundstart race, but it's a good way to test.


FlattestGuitar:
This won't fix it, but it's a good change, theoretically. Sure?

PR #638, opened 2018-02-17
Reenables plasmamen

Wilchenx:
I've seen no actual good reasoning why they were disabled

They're the most mechanically punishing and unique actual xeno race to play, everything else we have currently enabled is "animal but has human features" which is the opposite of the meaning of the word "Xeno" which means alien, as in something not unknown. Hybrids of earth species are not xeno, this on the other hand is real xeno beauty.

👏

PR #670, opened 2018-03-03
Return of the Plasmamen

Ziiro:
:cl: 
add: Plasmaman Lore Blurb
/:cl:

Followup to https://github.com/OracleStation/OracleStation/pull/638 . Adds a loreblurb, this will need a config update. I suggest 100 hours of gametime to play as a plasma man to keep them rare (and so people have a better understanding of mechanics before playing a race this dangerous.)

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/plasmamen.dm
+++ b/code/modules/mob/living/carbon/human/species_types/plasmamen.dm
@@ -22,6 +22,8 @@
 	ass_pic = "plasmaman"
 	examine_text = "a Plasmaman"
 	species_text_color = "#800064"
+	loreblurb = "The product of plasma experimentation on a human colony, these strange abominations lead a difficult life of breathing plasma and bursting into flames when exposed to oxygen. \
+	Logically speaking, they make for bad crewmembers. However, they don't have a choice but to work for the only corporation that can supply them with plasma."
 
 /datum/species/plasmaman/spec_life(mob/living/carbon/human/H)
 	var/datum/gas_mixture/environment = H.loc.return_air()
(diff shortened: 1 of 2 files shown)


ike709:
inb4 a negative comment from purp

FlattestGuitar:
I still think they're just too mundane to be added, but I won't oppose this if the community really wants it. 

More or less what the deal with Ethari was, we just don't have hydrocarbon furries.

Cactus2388:
Plasmaman still spawn from killing legion IIRC. Just wanted to point that out.

ike709:
@FlattestGuitar They're less mundane than Ethari, and I'm tired of the double standard in which we don't want reskin races, but we continue allowing Ethari to exist.

SmokingKilz:
It's not a matter of us letting Ethari exist for the sake of existing, it's more so there's no change that has been suggested that everyone agrees on. I still genuinely think making them wookies would be a great middle ground, but yeah.

AndrewMontagne (approved):
I don't mind re-adding plasmamemes, I'm just super wary of giving shitters the ability to run around the station on fire at will.

ike709:
@AndrewMontagne I think you underestimate just how weak plasmemes are to heat. I've seen rounds where literally all the plasmamen die because someone turned up the temperature on the distro loop gasses by a few degrees. Didn't affect the humans at all, killed the plasmemes in minutes.

Point is, they can't exactly run around on fire for long.

SamHPurp:
@Cactus2388 aye, thats fine though - even a good thing! This is more for round-start selection, than not available at all.

I've put this up to a vote with the maintainers.

SamHPurp:
We've agreed to test it for a week.

PR #622, opened 2018-02-14
Drone hatches

ike709:
:cl: Aurora, ported by ike709
add: Added drone hatches to most airlocks. Now maintenance drones shouldn't let people into places anymore.
/:cl:

I forgot to record audio, but here's some videos anyways from the FTL13 PR.
https://www.dropbox.com/s/d4li9ksrqgnpwdg/2017-10-15-0148-49.mp4?dl=0
https://www.dropbox.com/s/xeg5l3vuq4avu8l/2017-10-15-0149-37.mp4?dl=0

Tested and working.

Now if a drone is opening an airlock with a person nearby, you know it's to let them in/out!

PR #672, opened 2018-03-03
Revert "Drone hatches"

FlattestGuitar:
Reverts OracleStation/OracleStation#622

tfw


SamHPurp:
So this wasn't tested?

FlattestGuitar:
It was not, no.

ike709:
Actually I did test it. However, I only tested it as a drone...

PR #673, opened 2018-03-03
Drone hatches II: The Hatchinator

FlattestGuitar:
No changelog, because we already have this in ours.

Just a simple issue with CanPass not returning parent proc's value.


ike709:
```
/obj/machinery/door/CanPass(atom/movable/mover, turf/target)
	if(istype(mover))
		if(mover.checkpass(PASSGLASS))
			return !opacity
		if(density && has_hatch && mover.checkpass(PASSDOORHATCH))
			open_hatch(mover)
			return TRUE 
	return !density
```

ike709:
I don't know how I fucked up copypasting that from the FTL13 PR, but I did.

Ziiro:
Another quality PR by Ike

It was really funny to have the entire station screaming about their doors not working by the way

ike709:
To be fair I tested it with drones and it worked fine for them... I just didn't test it with humans.
@Ziiro

PR #680, opened 2018-03-06
Makes it so Admin frozen players can't be moved

Ziiro:
:cl: 
tweak: Makes it so Admin frozen players can't be moved
/:cl:

[why]: # People dragging admin frozen players is a problem. This fixes it.


Keluandrie:
I've dragged afrozens out of depressurized areas before, but I can see how shitters may cause this to be an issue.

scrubmcnoob:
Those situations are rare enough and admins can always rejuv.
This seems like a quality of life thing tbh.

Birdtalon:
You could always just make the person invulnerable anyway during the freeze in the proc itself if that is an issue.

AndrewMontagne (approved):
LGTM

PR #684, opened 2018-03-08
Refactor get_players_for_role

avosirenfal:
Helping @Ziiro try to track down the IPC changeling bug. Let's start with making this method readable.

This compiles, but I haven't tested it since you'd need more than one player anyway. That's up to you guys.

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -311,13 +311,35 @@
 		set_security_level(SEC_LEVEL_BLUE)
 
 
+/datum/game_mode/proc/is_player_eligible_for_role(role, mob/dead/new_player/player)
+	if(!player.client || player.ready != PLAYER_READY_TO_PLAY)
+		return FALSE
+
+	if(jobban_isbanned(player, "Syndicate") || jobban_isbanned(player, role))
+		return FALSE
+
+	if(!age_check(player.client))
+		return FALSE
+
+	if(restricted_species)
+		for(var/species in restricted_species)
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/game_mode.dm
```diff
@@ -326,91 +348,21 @@
...
-
-	return candidates		// Returns: The number of people who had the antagonist role set to yes, regardless of recomended_enemies, if that number is greater than recommended_enemies
-							//			recommended_enemies if the number of people with that role set to yes is less than recomended_enemies,
-							//			Less if there are not enough valid players in the game entirely to make recommended_enemies.
+				candidates_preferred += player.mind
+			else
+				candidates_nothanks += player.mind
 
+	if(candidates_preferred.len < recommended_enemies)
+		for(var/datum/mind/player in candidates_nothanks)
```
Birdtalon: Uh won't this start adding job-banned and ineligible players if there are not enough recommended enemies?

avosirenfal: No, ineligble players won't be in either list. `candidate_nothanks` just holds people who do not have antag enabled in their preferences (force antag). That's checked here- https://github.com/OracleStation/OracleStation/pull/684/files#diff-2334d1bad599b5c51def0eeaa65d7bfdR355

Birdtalon: Ah right I was reading the indentation incorrectly, Thought that if that proc returned false then it would add them to the `candidate_nothanks` list.

PR #685, opened 2018-03-08
Ziiro attempts to fix the IPC Cling problem again

Ziiro:
:cl: 
fix: No more IPC clings.. or are there?
/:cl:

https://github.com/OracleStation/OracleStation/pull/684 Is a way better idea. Go look at that PR.

This is a backup.

Thanks @deathride58 for this

🔫

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -336,16 +336,15 @@
 	if(restricted_species)
 		for(var/mob/dead/new_player/player in candidates)
 			for(var/species in restricted_species)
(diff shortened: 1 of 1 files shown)


FlattestGuitar:
But the problem is players not being removed from the pool, not the other way around.

These messages will literally just not print and give us nothing.

Ziiro:
heck

FlattestGuitar:
This would be the fix if this runtime.

This doesn't runtime, the player list already only has mobs with clients and all clients always have prefs, unless something fucks up really badly somewhere, which we'd be aware of otherwise.

I'm sorry, Ziiro, but dis is not da wae.

PR #690, opened 2018-03-10
Sleeper Agent/Double Agent Tweaks

Ziiro:
:cl: 
tweak: Change the Sleeper Agents' objectives to prohibit collateral damage.
balance: Sleeper Agents will now have lower numbers when present.
/:cl:

Less Murderbone
SLIGHTLY Less chaos

### DIFF
--- a/code/game/gamemodes/traitor/double_agents.dm
+++ b/code/game/gamemodes/traitor/double_agents.dm
@@ -3,7 +3,7 @@
 	var/list/late_joining_list = list()
 
 /datum/game_mode/traitor/sleeper_agent
-	name = "Sleeper Agents"
+	name = "sleeper agents"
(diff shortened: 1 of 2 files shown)


AffectedArc07:
No more plasafires :(

code/game/gamemodes/traitor/double_agents.dm
```diff
@@ -6,12 +6,12 @@
 	name = "Sleeper Agents"
```
ike709: "Sleeper Agents" is actually supposed to be lowercase. It was uppercase when it was called Internal Affairs so my find & replace kept it uppercase. All the other game mode names are lowercase.

If you could fix it for consistency that'd be great.

PR #701, opened 2018-03-11
Antag Playtime Requirements

FlattestGuitar:
Because holy heck, these 2-hour-wondernukies are starting to get annoying.

:cl: Flatty
add: Antagonists now have playtime requirements, play to unlock.
/:cl:

I haven't actually tested this beyond booting up a few things, so we should probably do that before a merge.


ike709:
I support the idea but I honestly think most of the times should be higher.

SamHPurp:
Will merge, but the actual numbers will be tweaked when updated in the config... so the live numbers aren't going to be these. I agree it should be higher.

SamHPurp:
Config change implemented. Numbers might need to be tweaked.

PR #702, opened 2018-03-11
/datum/oracle_ui

AndrewMontagne:
I've long talked about a replacement for TGUI, and here it is. I've refactored paper to use `/datum/oracle_ui` as an example on how to use it.

I've also created a subclass, `/datum/oracle_ui/themed` which supports templating and an accompanying tgui-like theme. With some simple clientside JS, I've managed to refactor the disposal bin to use `/datum/oracle_ui/themed/nano`, and the results are reasonably good.

gif

There is also an included web IDE which you can open and play with HTML and data. There's a helper proc called `oui_data_debug(mob/user)` you can invoke on any atom to get a data string to insert into the IDE.

image
image

### Summary

It does this:
- Wraps the `use << browse` functionality.
- Keeps track of all the people viewing the UI.
- Handles multicasting updates to viewers.
- Handles multicasting js invocations to viewers.
- Handles closing UI windows when players are no longer able to view them.
- Handles sending assets to viewers.
- Handles templating.
- Handles Topic() calls.

It does not do this:
- Use node.js
- Require complilation

### Done

- [X] Basic implementation.
- [x] Add stub methods to various classes?
- [X] Handle `Topic()` calls and URLs?
- [x] Add flashing to show updates.
- [x] Focus theft for windows.
- [x] Subclass which has templating from HTML files.
- [x] Basic theming support.
- [X] Fully flesh out the `nano` theme.
- [X] Decent documentation on how to use it.
- [X] Interactive editor for development?


FlattestGuitar:
I'd call it AndrewUI

html/simple_ui/themes/nano/fonts/SIL-OFL-1.1-LICENSE.md
```diff
@@ -0,0 +1,92 @@
+This license governs the Font Awesome font files located in tgui/assets/fonts
```
ike709: You should probably change that path.

ike709:
pic of disposal unit UI?

ike709:
nevermind I saw the one on discord:
https://cdn.discordapp.com/attachments/359016807501266944/423625349923078155/GIF.gif

html/simple_ui/themes/nano/fonts/SIL-OFL-1.1-LICENSE.md
AndrewMontagne: It was a copypaste of a folder, probably gonna remove it and cover the licensing in the README.md in the module folder.

code/modules/simple_ui/simple_ui.dm
```diff
@@ -0,0 +1,101 @@
...
+	viewers |= target
+	if(!isprocessing && (auto_refresh | auto_check_view))
+		START_PROCESSING(SSobj, src) //Start processing to poll for viewability
+	//Send the content
+	target << browse(get_content(target), "window=[window_id];size=[width]x[height];can_close=[can_close];can_minimize=[can_minimize];can_resize=[can_resize];titlebar=[titlebar];")
+
+/datum/simple_ui/proc/update_all()
+	for(var/viewer in viewers)
+		//Spawn forks off the execution to their own coroutines for better performance, since winget is a blocking call
+		spawn() render(viewer, TRUE)
```
MrStonedOne: Making render a non-blocking proc with `set waitfor = FALSE` is more preformant then spawning in a loop. as this creates 2 trips thru the sleep queue, one for the spawn, another for the winget, where as `set waitfor = FALSE` only has the winget sleep. Spawn also has to copy the current frame of the call stack where as `set waitfor = FALSE` just detaches the current frame.

AndrewMontagne: @MrStonedOne Thanks for that, I didn't know you could set a method as async that way. I'll refactor that after work tonight. 👍

AndrewMontagne:
`/datum/simple_ui` is functionally complete at this point. Just some polish and future-proofing to go now.

MrStonedOne:
>Interactive editor for development?

I've been idlily making a server side templating engine and have already made a basic live IDE for it

https://github.com/MrStonedOne/dmtemplate/tree/ide_demo

https://www.youtube.com/watch?v=k4qm4WvtG0U

AndrewMontagne:
Added a basic how-to and full API reference under [README.md](https://github.com/AndrewMontagne/OracleStation/blob/9eb76b28f8e8c74197bdb6014e029e87857cec26/code/modules/oracle_ui/README.md) in the oracle_ui module.

AffectedArc07:
I have some questions
- Will this UI have the colored eye thing to see if you are in clicking, viewing or out of range
- Will this be able to have colored progress bars and buttons
- Does it auto refresh without the need for spamming

AndrewMontagne:
@AffectedArc07 
- No, the UI closes when you go out of interact range.
- Yes, you can even define entirely custom stylesheets
- Yes, either the UI can poll the object or the object can push changes as soon as they happen.

PJB3005:
So, why not just bring Nano back? What does this bring to the table? I skimmed over it and it just seems like a weaker Nano with progress bar animations.

FlattestGuitar:
It completely eliminates the need to depend on any additional tools and is far easier to work with.

Considering 99% of the UIs out there don't need more than basic HTML and the progress bars, it's a pretty good way to de-clutter the whole thing.

SamHPurp:
Not needing it compiled / node is nice.

Its weakness is state-saving. IE: Writing something, and having it still be there w/ a dom-refresh.

So its faster and easier for many usecases, but certainly not all.

PJB3005:
> It completely eliminates the need to depend on any additional tools and is far easier to work with.

> Not needing it compiled / node is nice.

Same for Nano.

> Its weakness is state-saving. IE: Writing something, and having it still be there w/ a dom-refresh.

Same for Nano, but due to its [*incredible* flexibility](https://changelog.ss13.moe/static/ShareX/2017/06/2017-06-03_18-28-30.mp4) this is less of a problem.

PR #703, opened 2018-03-12
Syndie Bag/Mag Update And Fix

scrubmcnoob:
The syndicate is about murder, not pacifism.
The syndicate shotgun ammo bag now has 6 buckshot magazines rather than 6 tasers. Makes the previous one buckshot into a taser magazine.
Loading a dragonsbreathe into bulldog didn't appear, making it seem unloaded. Fixed.
Slug magazines now have a new sprite so it's easier for a person to quickly tell what is in there.
Dropped slug from buckshot slug in the name cause what the hell are buckshot slugs.

:cl: scrubmcnoob
add: Bulldog slug magazines have their own unique sprite.
tweak: Bulldog duffel bag has more buckshot than taser magazines.
fix: Dragonsbreath magazines now have a sprite while loaded in the gun.
/:cl:


FlattestGuitar:
I disagree with the change regarding taser boolets. I'd like to hear the arguments for the change, as I believe the current loadout supports more diverse approaches to shooting shit and is useful regardless of what you want to do with the shotty.

scrubmcnoob:
If people want a taser, they would grab a taser not a bull dog shotgun. 
The shotgun bag should specialize in buckshot, the generic and primary round type you think of when you hear shotgun
And what more diverse approach? Using a bulldog means you're most likely going loud as you're not being stealthy. Just tasing people down then clubbing them to death seems odd when they can just be doing the same job with buckshot.
Also the bag load out here is useful for 'shooting shit' too, and I say even more useful regarding what you want to do with a shotgun. Generally, you think of lead spreading with a shotgun not taser slugs.

SmokingKilz:
I don't see the point of having the taser slugs to begin with honestly, sure in some situations they could come in handy, but definitely not warranted to have a bag filled with 'em.

Ralta:
Completely agree with scrub and smoking on this one - nukies are a murdery antag and while they may have need of tasers occasionally (hostages, interrogation etc) they certainly don't need mags full of 'em.

... that said, I'd LOVE to see a nuke team war ops and succeed without killing anyone :P

PR #705, opened 2018-03-13
Tweaks to brain damage messages

FlattestGuitar:
As a similar PR back on Paradise once framed it, the brain damage messages have passed the Turing Test after I tried to question a person on their OOC IC. I think it's for the best to remove the ones that are a bit too OOC.

:cl: Flatty
del: Removed some of the more OOC-y brain damage messages. Probably for the best.
/:cl:

Discuss.


AffectedArc07:
kek

Ziiro:
We should probably add a bunch more to this list

Ralta:
Yeah I'm not a fan of the ick ock looking ones. Ive been fooled a couple of times but managed to catch myself before bwoinking. Still, they don't really fit the vibe of the server and kinda mess with immersion imo.

Sikroth:
I like the change. Probably should add a few new lines to have variety, I think.
Also this line should probably be removed too (I think it hasn't, but maybe I'm too dumb to read github changes):
         "@pick(bug)",

FlattestGuitar:
@Sikroth it indeed should be removed! Cheers.

Also, in the future:

PR #714, opened 2018-03-16
Messenger Bags

FlattestGuitar:
Ports Messenger bags from Bay.

You can pick those as an alternative to backpacks, satchels and duffels in the character setup menu.
They do not differ from the backpack in terms of functionality, pure fluff.

:cl: Flatty
add: Adds messenger bags! Pick those as an alternative to your backpack in the character setup menu.
/:cl:


Ralta:
Oo, these like nice. Want!

Ziiro:
Wait, so these cause the same slowdown as backpacks? They aren't like stachels?

Cactus2388:
> They do not differ from the backpack in terms of functionality, pure fluff.

@Ziiro

ike709:
They should be a middle ground between satchels and backpacks in terms of slowdown and capacity imo.

PR #716, opened 2018-03-16
That Medical Rework

FlattestGuitar:
New PR because it was a total mess.

## Sprites!
| Body Scanner  | Robot Analyzer | Fracture HUD Thingy |
| ------------- | ------------- | ------------- |
|   |   |  |

## Fracture HUD thingy Sprite:

# Medical Rework
In general, I've tried to change and tweak stuff so that Medbay offers a more interesting medical system and so that each casualty that ends up at Medbay is more unique.
## Organs and internal damage
Organs don't change much in function, they still do all their stuff, but now you can damage those organs. Sharp implements approaching spessmen at high velocities will contribute. Obviously, there is also organ repair surgery, for both robotic and squishy internals.
## Combat
There's now a bit more nuance to picking your weapons in combat. Sharp items will have a higher chance to stab into an organ and deal heavy internal damage, while blunt weapons (everything that isn't sharp) will have a much higher chance of breaking a bone.

### Meta:
Replaces a Medical Records Console

### Delta:
Replaces a potted plant 

### Omega:
Replaces a potted plant

### Pubby:
Replaces nothing

### Box:
Replaces a Medical Records Console 

I've also added Robot Analyzers to all the robotic labs on these stations. They analyze robotic bodyparts and organs, as well as let you check the integrity of Cyborgs and Bots. Not all that useful for the latter ones.

## TL;DR
* Organs can be damaged
* That damage can make an organ function not-quite-right
* Blunt weapons break bones more
* Sharp weapons break internal organs more
* Mitocholide is a chemical chemistry can make. Pretty challenging to produce and easy to overdose.
* Mining epipens have enough of mitocholide to heal 50 damage on every organ.

### The lovely scanner sprites from ERIS, the code for them heavily inspired by Paradise Station's implementation.


code/modules/surgery/organ_manipulation.dm
```diff
@@ -27,14 +27,18 @@
 	implements = list(/obj/item/organ = 100, /obj/item/reagent_containers/food/snacks/organ = 0, /obj/item/organ_storage = 100, /obj/item/device/mmi = 100)
 	var/implements_extract = list(/obj/item/hemostat = 100, /obj/item/crowbar = 55)
 	var/implements_mend = list(/obj/item/cautery = 100, /obj/item/weldingtool = 70, /obj/item/lighter = 45, /obj/item/match = 20)
+	var/implements_heal = list(/obj/item/stack/medical/bruise_pack = 90, /obj/item/stack/medical/gauze = 55, /obj/item/stack/medical/gauze/improvised = 35)
+	var/implements_heal_robotic = list(/obj/item/stack/nanopaste = 100, /obj/item/screwdriver = 30)
```
Ziiro: Screwdriver should probably be upped from 30 to 50. Or nanopaste would need to be provided in robotics. This incorrectly assumes that R&D actually does their job and/or supplies items to anyone outside of mining.

code/modules/mob/living/living_defines.dm
```diff
@@ -15,7 +15,7 @@
 	var/toxloss = 0		//Toxic damage caused by being poisoned or radiated
 	var/fireloss = 0	//Burn damage caused by being way too hot, too cold or burnt.
 	var/cloneloss = 0	//Damage caused by being cloned or ejected from the cloner early. slimes also deal cloneloss damage to victims
-	var/brainloss = 0	//'Retardation' damage caused by someone hitting you in the head with a bible or being infected with brainrot.
+	//brainloss moved to be handled on the brain organ itself. Rip.
```
Ziiro: This might be a dumb question. If you've moved this: does brainloss show up in the top left of VV still? Or did you remove that too? I've looked over this, and I'm probably missing it, but

AdamElTablawy:
How fast do you bleed out? Because I can see this making heparin incredibly strong.

ike709:
Alright everybody, let's have a quick little lesson about the basics of blood loss in real life. First, there are three types of bleeding: arterial, venous, and capillary. (Note that this lesson assumes you don't have hemophilia, which would be a great trait if we port those). 

- Capillary bleeding is basically harmless. It's the small cuts that you *might* feel inclined to put a bandaid on. 
- Venous bleeding is a bit more serious. You're bleeding from one of your veins. You can tell because the blood will ooze out, and is usually darker due to lack of oxygen. Five minutes of pressure followed by a bandage and it'll stop in most cases.
- Arterial bleeding is the most dangerous, because it's the hardest to stop. It comes from one of your arteries. You can tell it's arterial because the blood will spurt out, not ooze. Fifteen minutes of pressure, bandaging, and potentially sutures or tape (depending on severity of the wound) and you'll be fine. If it's a major artery, you'll probably need a tourniquet, which has its own problems.

Real life wounds are nothing like in the movies. You know in a movie where a guy gets shot and all of a sudden he bleeds everywhere and dies of blood loss? That doesn't happen in real life. In real life, as long as the bullet misses any major arteries or organs, you're going to be fine.

Point is, two stetchkin shots to the chest shouldn't cause someone to bleed to death in two minutes, assuming nothing important is hit. If you want to be realistic, shots to the chest should have a chance to hit an organ. If it doesn't, they should be more or less fine (Physically, at least. Do we have pain? Also, shock is still a very real possibility, but I don't think we have that.). If it *does* hit an organ, it should cause massive organ damage and bleeding that requires more or less immediate surgery to treat or they'll die in minutes. You can and should also give sharp objects attacking any limb the chance to cut a major artery, which will cause them to bleed out if a tourniquet and/or pressure isn't applied.

AdamElTablawy:
we aren't aiming for realistic this is a shitty spaceman video game where a wizard can name himself ENORMOUS PENIS and turn into a metal rod that penetrates things

get your realism arguments out of here

code/modules/surgery/organ_manipulation.dm
FlattestGuitar: Makes for an exciting Robotics visit!

FlattestGuitar:
Well, yeah, that's my main counterpoint to that. We're not trying to be realistic and only medical students will be able to point out that there's something wrong with the bleeding.

ike709:
@FlattestGuitar guess who's a medical student

oversane:
> Toxins break all your organs instead of being just a number

How am I supposed to eat as a Vox?

Vivalas:
I like everything here except not passing out at 50 oxyloss. Maybe make it a bit more like 80-100? Passing out due to lack of oxygen seems perfectly in place and relevant to a space station death simulator.

FlattestGuitar:
Ready for a proper review. The whole thing would definitely use that.

SamHPurp:
This will break Ashwalkers entirely. You'll need to come back with some considerations for them.

Ziiro:
Give lizards some kind of sleeper reskin at their base. Or simply roll with the how disposable they are even harder. 1 corpse = 1 egg.

FlattestGuitar:
@Ziiro the toxin change is no longer a part of the PR. You eat as a Vox exactly the same way as you did before. Only breathing space will damage your lungs. The wrong gas will only deal toxin damage.

Cactus2388:
This'll make mining 10x harder. Killing megafauna is already high enough. I recommend adding an immovable autosurgeon table that fixes bones and shit be added to the mining outpost or something

SamHPurp:
Lets get this play tested for at least 6 hours or so to catch all the bugs.

Ralta:
Messed around with this a bunch yesterday at CentComm. Couple minor bugs and some balancing could be done I think:

```Body scanner window doesn't close if you get into it. 

Body scanner doesn't show limbs that have been removed (for obvious reasons). 

Robot analyser doesn't do...anything?

Blood loss and bone factures need to be increased. 

Breaking bones should have higher chance at higher levels of damage, ideally. ```

The scanner UI thing is a small bug I assume, and while not exactly game-breaking, it's a bit silly. As for the limb removal, the limb just disappears from the list. It'd be nice if there was some way to highlight missing limbs or additional organs (eg alien embryo).

The Robot Analyser, when used, shows:

```Ralta is scanning BorgNameHere
Scanning: BorgNameHere```

And then no further information appears. Tried this with a few borgs, both player controlled and not. Inflicted damage and made them healthy, but still didn't show any detail. 

Finally, I think the rates of bone breaking and bleeding need to be increased. 

For bone breaks, I was able to bludgeon someone in the arm til it was at 100 damage and it didn't break. Sometimes it'd break after around 20 damage, but often I could go on hitting it over and over. I used a few common weapons such as wrench, toolbox, stun baton and telebaton, as well as an ablative baseball bat. The amount of damage inflicted didn't appear to change the regularity of bone breaks, but I can't say for sure. It'd be good if there was a threshold (say 50 damage) after which the bone is MUCH more likely to break. 

As for bleeding, I used a variety of weapons to cause bleeds, and felt the rate at which blood decreased was just far too low. Even after taking over 100 damage with internal organ damage on top, the patient was losing a percent every few seconds. Not sure what to suggest for this particularly, aside from a slight increase to blood loss in general. There's probably a more nuanced way of doing it involving amount of damage received, or 'types' of bleeding but I haven't thought of a decent way yet. 

Otherwise, everything seems to work just great; internal damage functions as I would expect, and shows up in he scanner correctly. Non-syndicate implants show in the scanner, as do various extra organs and such. Very pleased with it so far - good work, Flattest! :-)

AndrewMontagne:
Wiki Updated

PR #730, opened 2018-03-18
Tables don't stun

scrubmcnoob:
You no longer get stunned when seeing your ugly mug reflected on the table surface.

:cl: scrubmcnoob
tweak: Climbing tables no longer stun.
/:cl:

### DIFF
--- a/code/game/objects/structures.dm
+++ b/code/game/objects/structures.dm
@@ -81,9 +81,6 @@
 				user.visible_message("<span class='warning'>[user] climbs onto [src].</span>", \
 									"<span class='notice'>You climb onto [src].</span>")
 				add_logs(user, src, "climbed onto")
-				if(climb_stun)
-					user.Stun(climb_stun)
-				. = 1
 			else
 				to_chat(user, "<span class='warning'>You fail to climb onto [src].</span>")
 	structureclimber = null


ike709:
Without at least a short stun, there's no reaction time for the person on the other side of the table.

DaveTheHeadcrab:
Bad change in my opinion, being forced to drop your stuff and take a stun for a second or two is definitely a good feature. Defenders should definitely have the advantage here.

scrubmcnoob:
You already have a timer for climbing a table, and the defender can simply click on the table which stuns the user by knocking them over. I can always up the climb timer if people feel they need it.
Why should there be an additional stun? That is silly.

ike709:
Because you can't melee them until they're actually on the table.

scrubmcnoob:
Why would you need to melee them? Clicking the table simply knocks the person off which then stuns them for a bit.
Why should there be an additional stun timer when someone has already pass the check for climbing the table as anyone can interrupt them by running into them or clicking the table.

ike709:
Because beating the shit out of them is a better deterrent than repeatedly blocking their climb?

scrubmcnoob:
Call security? Shoot him with the many beanbag shells you get and throw him out? You don't need to beat the shit out of him
It's also an inconsistent feature I feel and out of place. You don't get stunned getting into a mech for example if you pass the timer check of climbing into it.
Also for gunfights, it means a defender doesn't even need to be near the table to defend the point and has to just wait for the attacker to get on, then fill them full of lead as their randomly stuck on the table.
The table can be defended , if you're there to push them off. If no one is there, anyone climbing shouldn't suffer a random stun penalty.

AdamElTablawy:
>You don't need to beat the shit out of him

there's a reason tables are used in clock cult defenses, dude. You're looking at this from the perspective of one nonantagonist climbing a table to reach another nonantagonist, not a hostile situation.

ike709:
@scrubmcnoob You're making several incorrect assumptions.

1. Sec exists.
2. Sec is competent.
3. Sec will respond in a timely manner.
4. Everyone has a shotgun like the bartender.

SmokingKilz:
I don't understand the issue of removing this honestly, I don't get why I'm going through a timer, getting stunned and on top of that dropping any item in my hand. Not only does it not make a lot of sense mechanically, it also simply doesn't make sense in general. If people LOVE making it really hard to climb a table in a pinch, then remove the stun but keep dropping an item.

AffectedArc07:
I for one like this

Raze500:
Just no, Then How i'm going to abuse those nerds? 

> Nerd
-A.M.I.G.O

SamHPurp:
I don't think this is a good idea.

Ralta:
I like the idea (dropping items in particular is annoying as fuck) but I think it'd need a test to see how it fares for things like clock cult.

scrubmcnoob:
Is anyone willing to give a concrete reason as to why they don't want this?
The game should not be promoting random stuns as this game being heavily stun based means one stun could result in your round in being over.
I can't think of doing any other action similar to climbing a table that will hand out a random stun after already passing the timer check.

Imagine sec rounding a corner and seeing an antag climbing a table to escape(assume I upped the climbing time by x1.5) The antag can either risk being tased by continuing to do their time check for climbing the table or simply disengage from the action and then can continue to either flee security with another route or fight sec. This is counter play, a choice for the player.
Imagine without this PR, as soon as sec rounded the corner the antag is on the table and stunned. The antag can't do an counterplay besides smash their keyboard in frustration as a stun keeps them in place for an easy taser shot.

Cactus2388:
With our current move speed I think removing the stun from climbing a table would be a nice change to add to the flow of gameplay

AdamElTablawy:
Don't climb a fucking table while security is chasing you, then. It's that simple. Tables should stun because it gives people on the other side of the table a chance to fuck the guy up if they climb on. That's it.

SmokingKilz:
It already gives the person on the other side a chance, all you have to do is click the damned table to knock them straight off. Why exactly would climbing a table even stun you, like what is the logic behind that.

SamHPurp:
Defenders should keep an advantage. And to those of you say "oh just click the table" - its really fucking easy to miss that someone is climbing a table, or if they haven't just covered the table with things already, so you can't even click it.

This actually benefits greytides, and Lord knows we've had more than enough of that.

Cactus2388:
> wow I actually have to pay attention to people climbing my tables now?

scrubmcnoob:
Then we up the climb time, make the text red and bigger? The defender can already push people off tables, no need for another stun.

SamHPurp:
Uh no, if I'm standing in front of the bar. You don't know which of the three tables I'm climbing onto. You have to click all three, and quickly... to stop me.

AdamElTablawy:
@SmokingKilz 

Ah yes, I'm going to click a table when i'm behind three other tables. The entire idea of 'table spam defenses' is that it's not literally one line of tables and you're not behind the first line.

Aside from that, it's easy to miss someone vaulting over a table. Just leave it as is.

SmokingKilz:
I'm personally still highly against the idea that climbing a table stuns you AND makes you drop what you're holding, I don't see the logic honestly.

SamHPurp:
Sorry, but no.

This is a key part of defense, and as it should, falls on the side on the defender.

Its really easy to miss text of someone climbing, as there is no visual aid. You also don't know which table they are on, so 'click the table' isn't a valid excuse when they could click any of them.

Not to mention how this interacts with clock-cult base defense already....

All this culminates in the fact that lots of other things are designed around this, and removing this is ultimately a bad thing for gameplay.

PR #739, opened 2018-03-19
New latejoin selection screen

FlattestGuitar:
Speaks for itself.
Feedback welcome.

Mostly from Paradise. Can't be bothered to look for the PR/commit.

:cl: Flatty
add: Fancy new latejoin Job Selection Screen
/:cl:

No, this is not made with AndrewUI.
No, I will not stop calling it AndrewUI.

### DIFF
--- a/code/modules/jobs/jobs.dm
+++ b/code/modules/jobs/jobs.dm
@@ -19,7 +19,8 @@ GLOBAL_LIST_INIT(medical_positions, list(
 	"Medical Doctor",
 	"Geneticist",
 	"Virologist",
-	"Chemist"))
+	"Chemist",
+	"Paramedic"))
 
 
 GLOBAL_LIST_INIT(science_positions, list(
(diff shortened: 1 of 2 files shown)


Ziiro:
`>clown isn't flagged as high priority`

Quantum-M:
I think the AI job should be bolded, and the HoP one seems to be missing (unless it's taken).

Looks very neat and nice though.

FlattestGuitar:
HoP is indeed taken on that pic. I couldn't be bothered to varedit the priority jobs.
Also, AI is not a command position, and that's the only reasoning I used for the boldening. It's technically snowflaking to make it bold, so I don't condone this.

SmokingKilz:
Looks great, well done my g.

SamHPurp:
Looks good.

But tbh, we should be robustly testing OracleUI #702 and moving everything to that, rather than developing and merging PRs that use nano

AndrewMontagne:
@SamHPurp This does not use nano/tgui, fyi.

Ralta:
It does look really nice. +1

AffectedArc07:
On a sidenote, does AndrewUI support the coloring of buttons, text, progress bars

PR #745, opened 2018-03-21
Tweaked MetaStation genpop.

AndrewMontagne:
image

- Removed windows
- Added cryo
- Added more lockers
- Added disposals bin
- Added second set of turnstiles

:cl: AndrewMontagne
tweak: MetaStation genpop is now a wee bit more greytide-resistant.
/:cl:

### DIFF
--- a/_maps/map_files/MetaStation/MetaStation.dmm
+++ b/_maps/map_files/MetaStation/MetaStation.dmm
@@ -210,11 +210,6 @@
 /turf/open/floor/plasteel,
 /area/security/prison)
 "aaB" = (
-/obj/structure/cable/yellow{
-	d1 = 1;
-	d2 = 2;
-	icon_state = "1-2"
-	},
 /obj/item/twohanded/required/kirbyplants{
 	icon_state = "plant-03";
 	layer = 4.1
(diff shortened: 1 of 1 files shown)


AffectedArc07:
MUH BEDSHEET CAPES

Quantum-M:
I think the lack of windows would affect antags a bit negatively, making it a bit harder for them to breaking out of permabrig via space.

The disposal chute is in a bit odd place, I say move it to where that potted plant is at, in front of execution.

I also think there should be a camera outside of permabrig.

AndrewMontagne:
@Quantum-M It's easy to escape out of genpop as a perma'd antag. Wait for someone with a non-perma sentence, steal their ID when their time is up, walk out.

scrubmcnoob:
Yeah, those windows were killing the prisoners. Also glad we'll have more lockers
>breaks windows
>gets electrocuted
>air spam locks them in and kills.

Ralta:
I love all of it except that disposal chute. I assume it's placed there so that prisoners can get out of the Brig which is fine and all, but it's kinda inconvenient right there.

Maybe stick it where the new upper right locker is or something?

SmokingKilz:
Possible fix to the no window argument is to have a single window up by where the garden is, that way if the garden is breached into only that place has no atmos?

AndrewMontagne:
@Ralta The disposals bin is for disposing of discarded personal effects in the locker room, since it tends to get cluttered rather quickly.

@SmokingKilz Nah. The most usual broken window is in the bathroom and that usually vents all of genpop anyway; best to keep no windows. Besides, they don't deserve such a nice view. =p

Xantholne:
Disposal bin looks awkward, I'd say move it down were the potted plant on the bottom left is.

FlattestGuitar:
DayZ said "YOLO".

PR #751, opened 2018-03-24
Some fixes or sth

### LINKED ISSUE #715: Combining item stacks is broken
What it says in the title. Sometimes adding a stack of items to another will only add 1 item to the resulting stack (e.g. If you combine two stacks of 10 you end up with 11 items instead of 20). I'm not sure if it happens when combining held items or when they combine on the floor, I haven't payed enough attention to that.

I've experienced this issue in a number of items. Namely, from the top of my head: mining beacons, floor tiles, metal rods, metal sheets, botany cloth/leather, sandstone bricks, etc. The easiest way to reproduce it I can think of it using the biogenerator in botany.

Steps to reproduce:
- Load and process some fruits in the biogenerator (at least 1000 biomass).
- Make a x10 cloth stack.
- Without picking it up, make anoter x10 stack.
- The cloth stack on top of the biogenerator now has 11 items in it instead of 20.

### LINKED ISSUE #721: Pin removal is still a thing despite us ditching firing pins
Issue reported from Round ID: 1291 (Oracle Station | Medium RP)

image

FlattestGuitar:
Fixes #721 
Fixes #715 

:cl: Flatty
fix: Stacks combine properly when coming from the biogenerator.
fix: The Pin Removal crafting recipe has been removed, since we don't even use pins.
/:cl:

### DIFF
--- a/code/modules/crafting/recipes.dm
+++ b/code/modules/crafting/recipes.dm
@@ -10,17 +10,6 @@
 	var/category = CAT_NONE //where it shows up in the crafting UI
 	var/subcategory = CAT_NONE
 
-
-/datum/crafting_recipe/pin_removal
-	name = "Pin Removal"
-	result = /obj/item/gun
-	reqs = list(/obj/item/gun = 1)
-	parts = list(/obj/item/gun = 1)
-	tools = list(/obj/item/weldingtool, /obj/item/screwdriver, /obj/item/wirecutters)
-	time = 50
-	category = CAT_WEAPONRY
-	subcategory = CAT_WEAPON
-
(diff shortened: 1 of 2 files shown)


AffectedArc07:
>fix: The Pine Removal crafting recipe has been removed, since we don't even use pins.

***pine removal***

FlattestGuitar:
Rip 🌲

Quantum-M:
But muh pines...

In all seriousness, think they mean that old pin recipe that didn't get removed when the pins stuff was removed  a long time ago.

FlattestGuitar:
You think?

PR #759, opened 2018-03-26
Adds Keycard Auths to QM Offices on Box and Meta

### LINKED ISSUE #660: QM needs a keycard authentication device in his office
The QM office needs a keycard authentication device to allow the QM, as a head, to swipe for red alert and other situations.

Lorwp:
:cl: lorwp
add: The QM's office on Box and Metastation have Keycard Authentication Devices now
/:cl:

Fixes #660. Omega doesn't get one because the QM doesn't have a office, but i can easily add one if wanted.

Box:
boxstation

Meta:
metastation-manyoumusthaveareallyslowconnectiontoseetheseifeelbadforyouman

### DIFF
--- a/_maps/map_files/BoxStation/BoxStation.dmm
+++ b/_maps/map_files/BoxStation/BoxStation.dmm
@@ -28387,7 +28387,6 @@
 "bor" = (
 /obj/structure/table,
 /obj/item/clothing/gloves/color/latex,
-
 /obj/item/razor,
 /turf/open/floor/plasteel/black,
 /area/science/robotics/lab)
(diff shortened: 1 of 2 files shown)


Lorwp:
I’ve never really seen any QMs do anything substantial with their authority outside of cargo if there’s a QM in the first place. Still, we should at least give any the chance to be Command staff (and we can easily revert this later with anything making the QM a head if we end up demoting them)

SamHPurp:
I appreciate this PR, and it was definitely missed when we did the original test for QM's being head of staff.

However, following discussions, observations and just plain asking people, we've decided that the QM is being put back in his place as a psuedo head.... Just need to get it PR'd instead. Someone, please. ;)

PR #770, opened 2018-03-28
buffs sarin stun

AdamElTablawy:
sarin will now stun for 1 second, as opposed to 1 decisecond, which i suspect was an oversight

:cl: optional name here
add: sarin has been buffed so that the stun effect will now actually apply a stun instead of instantly vanishing

/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/toxin_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/toxin_reagents.dm
@@ -917,7 +917,7 @@
 			if(prob(10))
 				M.confused += 15
 			if(prob(15))
-				M.Stun(1)
+				M.Stun(10)
 				M.emote("scream")
 		if(30 to 60)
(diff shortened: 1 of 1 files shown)


monster860:
Yeah, those stun delays aren't in deciseconds, they're in mob ticks, of which I believe there are 2 in a second.

AdamElTablawy:
@monster860 Except the thing is that no.

In practice, sarin stuns for what amounts to a decisecond. Based off literally every other chem that has any form of stun effect, this isn't 10 seconds, this is 1.

FlattestGuitar:
Unironic sarin buff.

AdamElTablawy:
@FlattestGuitar unironic sarin fix tbh, sarin sucks as a deathchem. it's legitimately worthless because it takes  2 minutes to actually do anything useful and it's pretty easily outdone by almost anything else, lexorin for example

PR #799, opened 2018-04-01
Removes cat ears from the AutoDrobe

AndrewMontagne:
:cl: AndrewMontagne
del: Removed cat ears from the AutoDrobe
/:cl:


Ziiro:
But how will we get those /tg/ bagil players now? 🤔

Cactus2388:
if you want cat ears you should have to steal them from an ethari

+1

Ralta:
Yes. Yes to the max.

SmokingKilz:
Tbh I really don't care for the cat ears so I'm fine with keeping or removing them.

PR #801, opened 2018-04-01
Changerdanger!!!

FlattestGuitar:
:cl: Flatty
add: Changelings now have a sting that allows them to test whether another person is a changeling. The victim will be informed of the test if it is successful within 10 to 20 seconds.
add: Changelings can now get the objective to kill another changeling. CHANGER DANGER!
fix: Changeling sting messages no longer print twice.
/:cl:

I definitely am taking feedback on the lines of words I wrote there. I am *not* a poet. I can't even speak English! I'm not a native speaker! I shouldn't be here!

There is a 40% chance that a changeling will have this objective. It's in the "steal" objective slot.

Discuss.


code/game/gamemodes/changeling/changeling.dm
```diff
@@ -113,7 +113,14 @@ GLOBAL_LIST_INIT(slot2type, list("head" = /obj/item/clothing/head/changeling, "w
 	absorb_objective.gen_amount_goal(6, 8)
 	changeling.objectives += absorb_objective
 
-	if(prob(60))
+	if(prob(40))
+		var/datum/objective/assassinate/changeling_assassinate = new // ass-assinate another changeling.
+		changeling_assassinate.owner = changeling
+		changeling_assassinate.find_target_by_role("Changeling", TRUE)
+		changeling_assassinate.explanation_text += "They are known in the hivemind as [changeling_assassinate.target.changeling.changelingID]."
```
FlattestGuitar: This will print that fancy "Mr. Psi" thing or whatever.

scrubmcnoob:
maybe make lings stay dead if decapped?

Ralta:
This is the kind of change I was hoping for with Changelings - sounds like a good idea to me.

The English is fine btw - makes sense, has good grammar and fits the same style as the other cling messages.

armrha:
This is good. Changelings roam around in packs like they are beloved family members right now, so adding to their paranoia and giving an incentive to maybe not do that is good. I also think we should reduce the scaling with population: 1 changeling per 6 players is far too high, I put it here though it's just a config change: https://github.com/OracleStation/OracleStation/pull/809

Ziiro:
How does this work with Xenobio lings?

SmokingKilz:
I would really like to see more significant changes to our current changelings. The fact that they can simply destroy the gibber and crematorium is a serious issues that needs to be addressed. I'd like to see it be so if a lings head is removed, they stay dead.

PR #802, opened 2018-04-01
Delta BS Fix/Wizard No mirror

scrubmcnoob:
:cl:scrubmcnoob
tweak: Wizards no longer have their magic mirrors
fix: Blueshield's can open their own door to leave their own spawn area on Delta.
/:cl:


scrubmcnoob:
Thanks Dreammaker.

Ziiro:
Wait what? Why remove the magic mirror? Barring the cult golem (that was removed) it makes for interesting race options

SmokingKilz:
Because it adds more negatives than positives, for example Zombie, etc.

PR #807, opened 2018-04-02
Adds alternative sprite selection for borgs & updates mining drone icons

Vivinet:
This adds alternative borg sprites with pop-up menu to select a different icon for your borg. Currently, there's only an alternative for the medical, mining, engineering and the unselectable sec-borg. These weren't too hard to add, so more could easily be added in the future.

bingweeeebeepboop

Each added sprite includes an eye state for lights being on, along with changes for the clock-cult servant status.

Included in this is the changing of the mining drone to be more up to date - new look and matching color scheme for all mining robotic units.

Before
south-exporteast-exportsoutha-export

After
gasgasgassidewaysboifastboyattacks

🆑 Vivinet
imageadd: New sprites for borgs and minebot
/🆑


code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -284,11 +284,25 @@
...
 		/obj/item/clockwork/weapon/ratvarian_spear)
-	cyborg_base_icon = "medical"
 	moduleselect_icon = "medical"
 	feedback_key = "cyborg_medical"
 	can_be_pushed = FALSE
-	hat_offset = 3
+
+/obj/item/robot_module/medical/be_transformed_to(obj/item/robot_module/old_module)
+	var/mob/living/silicon/robot/R = loc
+	var/borg_icon = input(R, "Select an icon!", "Robot Icon", null) as null|anything in list("Classic", "Droid")
```
FlattestGuitar: So that `list("classic", "droid")` could be turned into a single variable for each cyborg type that has an associative list with the names and icon states of each version. That would be a way cleaner implementation.
```
var/list/cyborg_versions = list("Classic" = "medical", "Droid" = "medicaldroid")
... way later in the actual proc ...
if(cyborg_versions.len > 1)
    var/borg_icon = input(R, "Select an icon!", "Robot Icon", null) as null|anything in cyborg_versions```

You should be able to figure out the rest. I'd rather not have people have to code this stuff when they want to add a new cyborg version.

Vivinet:
Going to hold off on this PR until new/different cyborg modules are implemented which can utilize an alternative icon.

PR #809, opened 2018-04-02
Changed changeling_scaling_coefficient from 6 to 8 to combat ridiculous changeling numbers on high pop

armrha:
:cl: Armrha
experiment: Adjusting changeling scaling coefficient from 6 to 12 (Changelings = Pop / Scaling Coefficient).
/:cl:

[why]: We've had a huge increase in player population lately, and there has been a strong trend toward Changeling rounds becoming extremely non-challenging for the Changelings. When Changelings outnumber any single department, they can roll up to any group and just take it over by force. They don't even have to bother with stealth. Armblades out they can roll sec and sec can't keep them all down long enough to dust even one generally when you have 9 changelings. This change doubles the scaling coefficient for then Changeling mode and would be worth testing to see if it makes those rounds more fun. Changelings should always have to be stealthy.

### DIFF
--- a/config/game_options.txt
+++ b/config/game_options.txt
@@ -210,7 +210,7 @@ SHUTTLE_REFUEL_DELAY 12000
 ## Used as (Antagonists = Population / Coeff)
 ## Set to 0 to disable scaling and use default numbers instead.
 TRAITOR_SCALING_COEFF 6
-CHANGELING_SCALING_COEFF 12
+CHANGELING_SCALING_COEFF 8
 
 ## Variables calculate how number of open security officer positions will scale to population.
 ## Used as (Officers = Population / Coeff)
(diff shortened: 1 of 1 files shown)


swagbringer:
[http://i.imgur.com/RSk1V.jpg](url)

SamHPurp:
With #801 on the cards as well, giving them a chance to have to wreck each other too.... perhaps this drop could be weakened to 8 - what do you think, @armrha ?

armrha:
Definitely worth trying. 8 + #801 would be a significant enough impact to play for a week or two and see if it makes big changeling rounds more fun. If people still have a lot of complaints we can always raise it later.

The changeling game mode is heavily based off the sci fi tropes in John Carpenter's 'The Thing', and I think the most fun I've had in Changeling rounds are where there was some attempt at stealth, misdirection, and imitation. While nothing is going to change how things roll when the Changelings have taken over all the head of staff positions, these changes should help us get back to:

1) Changelings have a reason to not act too overtly and are nervous about being exposed

2) The crew has some chance of actually winning Changeling mode, right now most people just don't even seem to bother trying to fight the Changeling Squad.

PR #824, opened 2018-04-07
Adds Ambient Occlusion (REQUIRES 512, WILL TRIP THE 512 REQUIRED FLAG)

deathride58:
This is a feature I'm planning on sending over to TGstation once the feature freeze ends there. This feature can be enabled or disabled by individual clients via the game preferences menu. The preference is located right underneath parallax.
Judging from testmerges over at Citadel, this won't result in any major performance issues.

Before (ambient occlusion disabled):
dreamseeker_2018-04-07_00-10-52

After (ambient occlusion enabled):
dreamseeker_2018-04-07_00-10-32

Small warning: This will make the server refuse all connections from clients running BYOND versions lower than 512, as the usage of the filter() proc trips the DM compiler's 512 required flag. This has the positive side effect of making it very difficult for most ban evaders to actually evade.

:cl: deathride58
add: Added ambient occlusion. You can disable this in the game preferences menu.
/:cl:

### DIFF
--- a/code/__DEFINES/layers.dm
+++ b/code/__DEFINES/layers.dm
@@ -6,6 +6,7 @@
 #define PLANE_SPACE -95
 #define PLANE_SPACE_PARALLAX -90
 
+#define OPENTURF_PLANE -2
 #define GAME_PLANE -1
 #define BLACKNESS_PLANE 0 //To keep from conflicts with SEE_BLACKNESS internals
 //#define TURF_LAYER 2 //For easy recordkeeping; this is a byond define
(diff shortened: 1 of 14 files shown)


AffectedArc07:
Does travis know how to compile 512?

SmokingKilz:
I still don't know how I feel about this, I want to like it... But I don't know honestly.

FlattestGuitar:
I see no reason not to merge it as an opt-in once 512 is released as stable. Until then, this PR will have to remain closed. 

Sorry!

AndrewMontagne:
Re-opened because 512 soon!

oversane:
ok but when will we have it merged

PR #826, opened 2018-04-07
Bumps travis up to 512.1420

Lorwp:
So #824 can merge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ dist: trusty
 
 env:
     global:
-    - BYOND_MAJOR="511"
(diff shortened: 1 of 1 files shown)


FlattestGuitar:
I don't think there are any issues with us doing that. We've been 512 compatible for a while now.

Unless anyone has any counterpoints, I'll merge this one.

SamHPurp:
My memory was that there were still some issues. 🤔

AndrewMontagne:
I am not comfortable doing this. This will directly affect the binary that is used on the server and could cause issues. I would rather wait until 512 is released before using it.

PR #831, opened 2018-04-07
Allows office chairs to rotate

Vivalas:
Adds can_buckled_rotate var to /atom/movable and edits FaceAtom proc so that if you are buckled to something you will turn with the buckled object if can_buckled_rotate is 1. Only office chairs utilize this for now.

:cl: Vivalas
add: Office chairs now rotate when you interact with other objects for maximum immersion
/:cl:

[why]: # (Please add a short description [on the next line] of why you think these changes would benefit the game. If you can't justify it in words, it might not be worth adding:) 

It's a nice fluff features from HRP servers.

### DIFF
--- a/code/game/objects/buckling.dm
+++ b/code/game/objects/buckling.dm
@@ -7,6 +7,7 @@
 	var/list/mob/living/buckled_mobs = null //list()
 	var/max_buckled_mobs = 1
 	var/buckle_prevents_pull = FALSE
+	var/can_buckled_rotate = 0 //can people rotate when buckled to this object? eg: rolling chairs
 
 //Interaction
 /atom/movable/attack_hand(mob/living/user)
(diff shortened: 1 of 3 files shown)


code/game/objects/buckling.dm
```diff
@@ -7,6 +7,7 @@
 	var/list/mob/living/buckled_mobs = null //list()
 	var/max_buckled_mobs = 1
 	var/buckle_prevents_pull = FALSE
+	var/can_buckled_rotate = 0 //can people rotate when buckled to this object? eg: rolling chairs
```
SamHPurp: `FALSE`

code/game/objects/structures/beds_chairs/chair.dm
```diff
@@ -193,6 +193,7 @@
 	anchored = FALSE
 	buildstackamount = 5
 	item_chair = null
+	can_buckled_rotate = 1
```
SamHPurp: `TRUE`

PR #832, opened 2018-04-08
Adds inflatable barriers

Vivalas:
Ports inflatable barriers from Paradise, which are originally from BS12. I think they are better than the holosign thing we have currently and look pretty nice. I don't like any of the weird pseudoscience air blocking things like holosigns and fans, and this alternative also has some uses, like being a bit more restrictive of access and allowing the consturction of airlocks and staging areas to breach zones.

:cl: Vivalas
add: Adds inflatable barriers, they can be ordered by cargo for 800 points, and also come in EngiVends.
/:cl:

[why]: It's a nice physical alternative to the engi holosign, and also is better at demarcating areas that are restricted and keeping out other personnel.


Kokojos:
I support this with all my love. Those barriers are lovely when dealing with a breach.

SmokingKilz:
Agree, I don't see a reason to not add these, I think they're a rather nice addition.

code/game/objects/structures/inflatable.dm
```diff
@@ -0,0 +1,292 @@
...
+	anchored = TRUE
+	opacity = FALSE
+
+	icon = 'icons/obj/inflatable.dmi'
+	icon_state = "door_closed"
+
+	var/state = 0 //closed, 1 == open
+	var/isSwitchingStates = FALSE
+
+/obj/structure/inflatable/door/attack_ai(mob/user as mob) //those aren't machinery, they're just big fucking slabs of a mineral
```
FlattestGuitar: copypaste falling out of your pockets, bby >:C

code/game/objects/structures/inflatable.dm
```diff
@@ -0,0 +1,292 @@
...
+
+/obj/structure/inflatable/door/CanAtmosPass(turf/T)
+	return !density
+
+/obj/structure/inflatable/door/proc/TryToSwitchState(atom/user)
+	if(isSwitchingStates)
+		return
+	if(ismob(user))
+		var/mob/living/M = user
+		if(world.time - M.last_bumped <= 60)
```
FlattestGuitar: make that `6 SECONDS` for my OCD please

code/game/objects/structures/inflatable.dm
```diff
@@ -0,0 +1,292 @@
...
+/obj/structure/inflatable/door/proc/SwitchState()
+	if(state)
+		Close()
+	else
+		Open()
+	air_update_turf(TRUE)
+
+/obj/structure/inflatable/door/proc/Open()
+	isSwitchingStates = TRUE
+	//playsound(loc, 'sound/effects/stonedoor_openclose.ogg', 100, 1)
```
FlattestGuitar: rip these comments out please

code/game/objects/structures/inflatable.dm
```diff
@@ -0,0 +1,292 @@
...
+		src.transfer_fingerprints_to(R)
+		qdel(src)
+	else
+		visible_message("[src] slowly deflates.")
+		spawn(50)
+			var/obj/item/inflatable/R = new /obj/item/inflatable(loc)
+			src.transfer_fingerprints_to(R)
+			qdel(src)
+
+/obj/structure/inflatable/verb/hand_deflate()
```
FlattestGuitar: For optimal UX, I'd like it to be possible to deflate them with an Alt+click, with a complementary bit in the examine description.

code/game/objects/structures/inflatable.dm
Vivalas: I just ported it

code/game/objects/structures/inflatable.dm
Vivalas: 6 seconds is a macro?

code/game/objects/structures/inflatable.dm
Vivalas: >:C

Vivalas: I think there has been discussion / contention before about this, either on paradise or baystation. Making it an alt-click thing makes it a bit too easy IMO to deflate by accident and cause a pressure casualty, and consequently adds a fair bit of deniability as well for griefers.

I suppose I could lock it behind a prompt that follow the alt-click, but then... what's the point?

Vivalas: I mean, alt-click is the shortcut for viewing the floor tile, something a lot of people use, at least I do. I know I would probably accidentally deflate it a bunch of times if I added an alt-click thing. Maybe alt-double click but I don't know how the game supports that and it might become snowflakey for this one feature.

code/game/objects/structures/inflatable.dm
FlattestGuitar: yes

code/game/objects/structures/inflatable.dm
FlattestGuitar: A prompt would be an improvement over using the right-click menu. A lot of work has been done to move towards a more consistent and useful interface and that thing is not it.

FlattestGuitar:
I still fail to see the point. These are far inferior to the holobarriers we have now and it seems like adding complexity for the sake of it. 

@SamHPurp, @SmokingKilz, you guys still want it in? If so, I can make the edits myself and merge this.

Vivalas:
Almost all of SS13's systems are complexity for the sake of it. One could argue that your medical changes are complexity for the sake of it in the form of added medical labor.

The main reason I am doing this is that "fans" and "holobarriers" are terrible for suspension of disbelief when tied in to the rest of the atmospheric system which strives to be somewhat accurate. They also look nice and make construction zones look a bit more thematic. I'll get to making the changes tomorrow. Since I believe adding a popup is worse than having to right-click, I'll make a system that deflates it if you alt-click twice in a certain time frame.

FlattestGuitar:
Sounds good, @Vivalas. 

Also, it's more about creating challenges and interesting gameplay with more variables, which I don't think this really adds. 

As far as I can tell these are completely inferior to the holobarriers, but I won't oppose this if people really want it in there.

code/game/objects/structures/inflatable.dm
```diff
@@ -0,0 +1,295 @@
...
+	playsound(loc, 'sound/items/zip.ogg', 75, 1)
+	to_chat(user, "<span class='notice'>You inflate [src].</span>")
+	var/obj/structure/inflatable/R = new /obj/structure/inflatable(user.loc)
+	src.transfer_fingerprints_to(R)
+	R.add_fingerprint(user)
+	qdel(src)
+
+/obj/structure/inflatable
+	name = "inflatable wall"
+	desc = "An inflated membrane. Do not puncture. Alt-click twice in quick succession to deflate."
```
FlattestGuitar: Wrap that alt-click bit in `<span class='notice'>` and make the `Alt-click` bold. Put the entire thing on a new line.

PR #834, opened 2018-04-08
New syndie base

SamHPurp:
"Redesigned the syndicate lavaland base because I felt the current one was visually kind of uninteresting with its square, boxy shapes and neatly organized rooms. So I made it dark, dank, and a bit more stylish!

This also provides it with functioning atmospherics and power, as well as a turbine engine that people can muck around with (and possibly die to). Finally, there are a couple of UNUSED potential shuttle docks, in case someone (namely me) wants to allow shuttles to dock with this station, such as the whiteship or nuke op ship or something else I design.

The shuttle computer in cargo bay does not actually do anything or connect to any existing shuttle. It's just for show. And yes, all external airlocks have tiny fans to prevent atmos processing all the time as well as objects going off into the lava.

Overall this map is around 50x46 (I can't quite remember) whereas it was previously around 38x38."

Re-wired, re-piped, and re-tiled most of the original port... Added some more Oracle-bits too.

Images coming:

🆑 WJohnston & Purpose
add: Redesigned the entire syndicate lavaland base to be more stylish, have functioning power and atmos as well as a turbine to play with.
/🆑


SamHPurp:
@AndrewMontagne I didn't update the photos because I forgot. But the tiles are fixed.

SmokingKilz:
Looks great honestly, I think the syndicate base has a lot of potential for the future.

Ziiro:
Huge props, I had this on my port list, and I also had some of the related fixes on that. You might want to add a couple of these if you didn't see them at the time.

https://github.com/tgstation/tgstation/pull/34764
https://github.com/tgstation/tgstation/pull/34143
https://github.com/tgstation/tgstation/pull/34401

SamHPurp:
@Ziiro I did the 2nd one, and the third one I handled in a different way.

With all the generation I did to test it, I couldn't make the first one occur, I think it has to do with how they changed their proc gen stuff.

So in theory all of them are sorted, but the first one might need some looking at in the future.

Issue #787, opened 2018-03-31
Blood cult target group objective assignments


SamHPurp:
Apparently this has existed for ages but nobody got around to bug reporting it. This is a high priority problem.

Blood cult group sacrifice targets are getting randomly assigned to different people, when all blood cult should have the same target. This only happens mid-round, not at round-start (as far as I'm aware).

My assumption is that this is to do with our cryo-pod code, and is noting that the target cryo'd then looping all people who had them as an objective and assigning a new target at random - when actually it needs to be decided upon, then looped through and assigned.

My assumption could be false.

vverti:
Sorry for never reporting this earlier, one thing I will note is that it seems to be that whoever the newest convert is SEEMS (not certain) to have the correct sacrifice target.

 Last night something new popped up in this issue which was the fact that, despite the sacrifice target being non-SSD/catatonic or cryoing, cult members being told, and I quote "Nar'Sie murmurs, Keith Riker is beyond your reach. Sacrifice Catherine Small instead..." While ALL the cultists still had to sacrifice their original target in their objectives. 

This makes me think that for some reason it is randomly deciding the sacrifice targets are no longer available, and so it tries to give a new objective, but never actually GIVES the new objective. 

I will update if anything new comes to light.

PR #839, opened 2018-04-10
Cultist fixed

Bksparkie1:
:cl: Bksparkie1
tweak: tweaked the cryopod code for cultists so that it only notifies players if the person is a legitimate target
/:cl:

[why]: # Cultists currently get notified that their target changed every time someone uses a cryopod..


DaveTheHeadcrab:
1) This isn't the bug reported in #787, #787 reports that upon the cryoing of the actual sacrifice target everyone is assigned a seemingly random and different sacrifice target, instead of one shared target like the cult is supposed to receive.

2) This doesn't fix the bug you're trying to fix either. I'll leave a code review to explain what's wrong.

code/game/machinery/cryopod.dm
```diff
@@ -269,10 +269,10 @@
 					reshape.Crop(7,4,26,31)
 					reshape.Crop(-5,-3,26,30)
 					GLOB.sac_image = reshape
-
-					for(var/datum/mind/H in SSticker.mode.cult)
-						if(H.current)
-							to_chat(H.current, "<span class='danger'>Nar'Sie</span> murmurs, <span class='cultlarge'>[occupant] is beyond your reach. Sacrifice [GLOB.sac_mind.current] instead...</span></span>")
+					if(GLOB.sac_mind == occupant)
```
DaveTheHeadcrab: One, GLOB.sac_mind is a mind datum, occupant is a mob. These two will never be equal.

Two, even if they were of the same type, this statement happens after we pick a new target to be the sac_mind and the list of potential targets excludes the occupant, so the occupant could never be selected anyhow.

Three, in the current position even if the GLOB.sac_mind was not updated to the new target yet, all this would do is stop the new target from displaying.

code/game/machinery/cryopod.dm
```diff
@@ -269,10 +269,10 @@
...
+						for(var/datum/mind/H in SSticker.mode.cult)
+							if(H.current)
+								to_chat(H.current, "<span class='danger'>Nar'Sie</span> murmurs, <span class='cultlarge'>[occupant] is beyond your reach. Sacrifice [GLOB.sac_mind.current] instead...</span></span>")
```
FlattestGuitar: `GLOB.sac_mind` is already the new, picked target. You need to find a different way.

code/game/machinery/cryopod.dm
DaveTheHeadcrab: Move this up between lines 246 and 247, replace occupant with occupant.mind

SamHPurp:
@DaveTheHeadcrab thoughts?

DaveTheHeadcrab:
At a glance looks fine, between lectures and class right now so cannot confirm.

At worst it shouldn't fuck anything else up.

I hope.

PR #841, opened 2018-04-11
Malf verbs fixed

### LINKED ISSUE #825: malf AIs can use doomsday while dead
or any ability, seemingly. yeesh.

Bksparkie1:
:cl: Bksparkie1
add: a function call to remove Malf verbs on AI death
/:cl:

[why]: # Dead traitor AI could summon doomsday etc when they clearly should not be able to.
if I did everything right, this 
fixes #825


Bksparkie1:
remaking with fixed commits

PR #842, opened 2018-04-11
Chem Implant Action Button Removal

scrubmcnoob:
ports tgstation/tgstation#36021
:cl: Improvedname, ported by scrubmcnoob
fix: chem implants can no longer be self triggered
/:cl:

### DIFF
--- a/code/game/objects/items/implants/implant_chem.dm
+++ b/code/game/objects/items/implants/implant_chem.dm
@@ -4,6 +4,7 @@
 	icon_state = "reagents"
 	origin_tech = "materials=3;biotech=4"
 	container_type = OPENCONTAINER_1
+	activated = FALSE
 
 /obj/item/implant/chem/get_data()
 	var/dat = {"<b>Implant Specifications:</b><BR>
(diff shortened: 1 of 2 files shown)


AffectedArc07:
Disagree, there should be some risk in using these, similar how the tracking implant allows anyone to TP to the one with the implant.

AdamElTablawy:
@AffectedArc07 The issue is that chem implants are basically used as ghetto dental implants that don't need surgery for people who want to pop their combat mixes of meth and heal

AffectedArc07:
Adam, If they broke into the secure part of sec just for that, I think they should be allowed to pop it in combat for the effort they went through to get it

scrubmcnoob:
What if security just starts using these things for combat implants
Read the thread at TG. We already have a surgery procedure that does this.

swagbringer:
Good port, thankfully I have never seen people use this, but its absurdly strong.

SamHPurp:
I don't think this is controversial really.

This is sensible. These are designed to be remotely triggered.

AdamElTablawy:
@AffectedArc07 Antags do not break into sec for this.

The HoS or Captain go and give THEMSELVES combat stim chem implants.

code/game/objects/items/implants/implant_chem.dm
```diff
@@ -39,10 +39,6 @@
 		return 0
 	var/mob/living/carbon/R = imp_in
 	var/injectamount = null
-	if (cause == "action_button")
```
FlattestGuitar: You made it not inject anything ever, while also leaving the action button in place. 

I'm not sure how this passed on /tg/, maybe I'm missing something. Have you tested this PR?
:thinking:

FlattestGuitar:
@scrubmcnoob, please reply.

scrubmcnoob:
I'll look into it.

scrubmcnoob:
Alright, fixed using tgstation/tgstation#36087

PR #844, opened 2018-04-11
Fixed the AI (With no stupid additional commits)

### LINKED ISSUE #825: malf AIs can use doomsday while dead

Bksparkie1:
🆑 Bksparkie1
add: added a call for the function to remove malf verbs so the ai cannot access it after dying
/🆑
fixes #825

### DIFF
--- a/code/modules/mob/living/silicon/ai/death.dm
+++ b/code/modules/mob/living/silicon/ai/death.dm
@@ -20,7 +20,7 @@
 	SSshuttle.autoEvac()
 
 	ShutOffDoomsdayDevice()
-
+	malf_picker.remove_malf_verbs() 
 	if(explosive)
 		spawn(10)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/ai/death.dm
```diff
@@ -20,7 +20,7 @@
 	SSshuttle.autoEvac()
 
 	ShutOffDoomsdayDevice()
-
+	malf_picker.remove_malf_verbs()
```
FlattestGuitar: `malf_picker` is not going to be there for a regular AI, so it will try to call the proc of a datum that doesn't exist. Needs to check for its presence first, with 

```
if(malf_picker)
    malf_picker.remove_malf_verbs()
```

SamHPurp:
Indentation errors, failing TRAVIS

PR #846, opened 2018-04-11
Splits mutations Into Subfolder

scrubmcnoob:
Mimics tgstation/tgstation#33079
Not a  direct port


SamHPurp:
@scrubmcnoob Please detail the differences?....

scrubmcnoob:
@SamHPurp We had some own changes to genetic code and if I did a direct port, we would lose them.

PR #847, opened 2018-04-11
Admin Announce Tweak

### LINKED ISSUE #838: Admin announcements aren't very visible
This causes miscommunication and strong opinions.

Please upgrade the visibility of said announcement, or add a sound.
*disclaimer il probably do it soon myself

scrubmcnoob:
image
The admin announcement is no longer normal size text so it isn't missed.

Fixes #838

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -474,7 +474,7 @@
 	if(message)
 		if(!check_rights(R_SERVER,0))
 			message = adminscrub(message,500)
-		to_chat(world, "<font size='15' color='blue'><b>[usr.client.holder.fakekey ? "Administrator" : usr.key] Announces:</b><font size='4' color='black'> </span>\n \t [message]")
+		to_chat(world, "<span class='adminannouncement'><b>[usr.client.holder.fakekey ? "Administrator" : usr.key] Announces:</span><span class='adminnotice'>\n \t [message]</b></span>")
 		log_admin("Announce: [key_name(usr)] : [message]")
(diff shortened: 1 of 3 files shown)


Kokojos:
:+1:

code/modules/admin/admin.dm
```diff
@@ -474,7 +474,7 @@
 	if(message)
 		if(!check_rights(R_SERVER,0))
 			message = adminscrub(message,500)
-		to_chat(world, "<span class='adminnotice'><b>[usr.client.holder.fakekey ? "Administrator" : usr.key] Announces:</b></span>\n \t [message]")
+		to_chat(world, "<font size='15' color='blue'><b>[usr.client.holder.fakekey ? "Administrator" : usr.key] Announces:</b><font size='4' color='black'> </span>\n \t [message]")
```
FlattestGuitar: The `color` tags are outlawed. Please use spans. We're already overflowing with random bullshit which makes it harder to change all of these once we go through with a proper UI redesign. 

Look at `browserOutput.css` and `browserOutput.dm`. Both files need to have a new span added if you want to make a fresh one.

SamHPurp: @FlattestGuitar 
image

FlattestGuitar: I know. It only became apparent after I actually tried to redo the spans.

You were right. ;c

scrubmcnoob:
Alright, used proper tagging and it still has the same effect of being noticeable and big.

SamHPurp:
(this is obnoxiously large)

PR #871, opened 2018-04-29
Donator stuff

FlattestGuitar:
### OBLIGATORY PRE-PUBLIC-OUTRAGE STATEMENT:
* This will knock down everyone's character slot count from 8 to 3.
* This will disable the ability for people to change their ghost shapes and orbits.
-------
With that said, all of this is still VERY MUCH up in the air. More perks will come when we get around to implementing more systems like loadouts or job titles (not saying we will; we just might!) or whatever else might be out there, but for now this will be a thing. I'm sorry!

Also, the donation tiers are also so far undecided.

I mostly need a review on the SQL stuff.


990xan:
As someone who uses all the char slots, and as someone who likes the custom ghosts options, H E C K no.

SQL/tgstation_schema.sql
```diff
@@ -286,6 +286,7 @@ CREATE TABLE `player` (
   `lastadminrank` varchar(32) NOT NULL DEFAULT 'Player',
   `accountjoindate` DATE DEFAULT NULL,
   `flags` smallint(5) unsigned DEFAULT '0' NOT NULL,
+  `donator` tinyint(1) DEFAULT '0',
```
FlattestGuitar: I actually couldn't find any documentation on what the heck these `(numbers)` mean. It doesn't make sense for them to be bytes, because that's already governed by the tinyint bit, so I've no idea. I've never seen this before. Andrew halp

SQL/database_changelog.txt
```diff
@@ -1,10 +1,21 @@
...
+INSERT INTO `SS13_schema_revision` (`major`, `minor`) VALUES (3, 3);
+
+----------------------------------------------------
+
+28 April 2018, by FlattestGuitar
+
+Added the `donator` column to the `player` table
+
+ALTER TABLE `player`
+ADD COLUMN `donator` tinyint(1) DEFAULT '0' AFTER `flags`
```
FlattestGuitar: **SQL CHANGE**

code/modules/donator/donator.dm
```diff
@@ -0,0 +1,104 @@
...
+
+	usr << browse(output,"window=donatorperms;size=900x650")
+
+/datum/admins/proc/set_donator_status(new_ckey, new_status)
+	if(!usr.client)
+		return
+	if(!check_rights(R_PERMISSIONS))
+		return
+	new_ckey = sanitizeSQL(new_ckey)
+	var/datum/DBQuery/query_donator_rank_update = SSdbcore.NewQuery("UPDATE [format_table_name("player")] SET donator = '[new_status]' WHERE ckey = '[new_ckey]'")
```
FlattestGuitar: Not sure whether this SQL query is correct. Andrew please look at it to try to find anything wrong with it.

Thanks.

code/modules/donator/donator.dm
```diff
@@ -0,0 +1,104 @@
...
+GLOBAL_LIST_INIT(donator_tiers, list("Not a Donator" = 0, "Basic Donator" = 1, "Full Donator" = 2))
+GLOBAL_LIST_EMPTY(donators)
+
+/proc/load_donators()
+	if(!SSdbcore.Connect())
+		log_world("Failed to connect to database in load_donators(). No donators will be loaded.")
+		message_admins("Failed to connect to database in load_donators(). No donators will be loaded.")
+		return
+
+	var/datum/DBQuery/query_get_donators = SSdbcore.NewQuery("SELECT ckey donator FROM [format_table_name("player")] WHERE donator > 0;")
```
FlattestGuitar: Another SQL query which I'm not confident in.

ike709:
I offered to do loadouts multiple times and the idea always got shot down.

Cactus2388:
We need more than 3 character slots

-1 until changed

ShattForte:
Only thing I'd like to suggest is possibly donators not having to wait for job and/or race timers.

FlattestGuitar:
This will not happen.

PR #873, opened 2018-04-30
Ports species flags for viruses

Kokojos:
:cl: optional name here
add: Each race now has a tag specific to it, that allows/disallows it to get viruses. Specific viruses can affect specific types of races.
add: Some viruses now affect the dead.
/:cl:

[why]: # 
First step before adding more viruses and strains of viruses. New illness for everyone, including the dead, woo!

From : https://github.com/tgstation/tgstation/pull/32858/files
WARNING : This DOES NOT add the dme for this file : code/datums/diseases/advance/symptoms/species.dm because Git hates me and I hate it. Will fix eventually or a maintainer will do.


code/modules/mob/living/carbon/life.dm
```diff
@@ -239,7 +239,7 @@
 		if(prob(D.infectivity))
 			D.spread()
 
-		if(stat != DEAD)
+		if(stat != DEAD && !D.process_dead)
```
FlattestGuitar: shouldn't this be `(stat != DEAD || D.process_dead)`?

FlattestGuitar: Yes, it definitely should. /tg/ later changed it. https://github.com/tgstation/tgstation/pull/35618

PR #875, opened 2018-05-01
LOOC tweak

FlattestGuitar:
Ghosts will now print their ckey instead of their mob name, unless they have dchat anonimity turned on, in which case they will still print their mob name.

:cl: Flatty
tweak: Non-anonymous ghosts will now print their CKEY when talking in LOOC.
/:cl:


FlattestGuitar:
Different PR.

Issue #884, opened 2018-05-03
Putting meatball spaghetti in a tray makes it look like spaghetti


LuamTheElf:
Issue reported from Round ID: 2146 (Oracle Station | Medium RP)

How to replicate, put meatball spaghetti into a tray

PR #885, opened 2018-05-03
Adds new diseases and an event to go with it.

Kokojos:
:cl: New diseases!
tweak: The spontaneous appendicitis has been replaced with spontaneous illness, causing a variety of things.
add: Beepitis B is a virus affecting only IPCs
add: Leg cramps can affect all humanoids, including IPCs. Charlie Horses incomming
add: Meningitis causes head pains.
/:cl:

[why]: #884 
New diseases are fun and provide more variety than ''MY APPENDIX HALP'' Also, this will be the start of my work on diseases, so I will build unto this.


Ralta:
Couple of critiques:

We already have a 'disease outbreak' event, do we not? I know there's a separate spontaneous appendicitis event too, but I think these diseases should get added to the existing disease outbreak. Or if they're being categorised as more 'minor' diseases, perhaps spontaneous appendicitis could be changed to spontaneous minor illness or some such? 

Also, I don't think adding diseases for IPCs is such a good idea purely because we've been trying to make each species 'feel' at least a little different when played. IPCs are immune to all disease, as well as radiation and toxin damage, so adding a disease - even an IPC-specific one - would rather counteract this. I suppose you could get around this by calling it something else, as if to imply it's a 'manufacturing defect' or something? 🤔 

In any case, I'm always happy to see more interesting symptoms and such appear.

spoonitate:
considering the disease outbreak is preceded by a gigantic broadcast, I personally don't see anything wrong with minor inconvenient illnesses. 

Though to keep it in the spirit of spontaneous appendicitis, I personally think that any illnesses spawned from this event should remain non-contagious. Otherwise those contagious diseases could be moved to the level 7 viral biohazard event.

Tayyyyyyy:
You should port the disease antagonist! tgstation/tgstation#35988

Wilchenx:
Removing appendictis and adding 4 diseases that literally do not do anything is lame as fuck.

You could've thought of a better game mechanic than just taking a shower for the cure.

FlattestGuitar:
So, my stance on this:
I don't think the diseases are interesting enough to warrant being added to the game. Appendicitis, even while being horribly inconsequential, still forces some cooperation between your average crewmember and Medbay. Stuff like leg cramps only serve to annoy people.

I won't close this without one of the Heads agreeing with me, but I really don't see the point.

SmokingKilz:
While we appreciate the work you've done here I have to agree with flattest, if you're willing to make some changes to the viruses so they do things that are a little more interactive we can talk about re-open this discussion. Sorry pal, Flattest can go ahead with the close.

PR #889, opened 2018-05-05
hotfix for chem recipes requiring fungus until a way to get fungus is added

imsxz:
:cl:imsxz
fix: chem recipes requiring fungus have been replaced with mushroomhallucinogen until a way to get fungus has been added
/:cl: 
chems that can be very necessary during certain situations like spaceacillin cant be made currently because theres no way to get fungus, spaceacillin is needed to cure fungal TB. theres other important med chems too like LSD that needs fungus. this PR replaces fungus in recipes to mushroomhallucinogen until a way to get fungus is added.

also webeditor is great and anyone who says otherwise has a wrong opinion


FlattestGuitar:
webeditor sucks

PR #903, opened 2018-05-11
Ashwalker warning

Ziiro:
:cl: 
tweak: Ashwalkers now get a clearer warning on spawn to not leave lavaland.
/:cl:

Removing ambiguity

### DIFF
--- a/code/game/objects/structures/ghost_role_spawners.dm
+++ b/code/game/objects/structures/ghost_role_spawners.dm
@@ -44,12 +44,12 @@
 	anchored = FALSE
 	density = FALSE
 	flavour_text = "<font size=3><b>Y</b></font><b>ou are an ash walker. Your tribe worships <span class='danger'>the Necropolis</span>. The wastes are sacred ground, its monsters a blessed bounty. \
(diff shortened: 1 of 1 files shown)


Cactus2388:
requires the "I ded pls nerf" and "removal" tags

Ziiro:
Wouldn't be opposed. This was triggered by an Ashwalker becoming a zombie and getting onto the station to create more zombies.

Cactus2388:
If we genuinely feel ashwalkers shouldn't be on-station make it so they can't use computer consoles. If they manage to get onboard after that kudos to them.

Ziiro:
I think that justification should only come from Admin intervention

SmokingKilz:
That kind of goes against the point of keeping ghost roles as free as possible, the idea is if someone can RP it in a cohesive way and justify their actions, it's fine. But if Ashwalkers storm the station and try to delam the SM "cuz ooga booga" then it's not okay.

FlattestGuitar:
Yeah, I gotta agree with Smoking. I'd put more pressure on the fact that Ash Walkers should defend the tendril, but also give them more room to do stuff. Also, softly pushing them towards something other than violent murder might not be a terrible idea.

PR #946, opened 2018-06-04
Bible Sounds Revisited

Affurit:
🆑
soundadd: Heals from the bible play an appropriate sound.
/🆑

Holy Healing now plays a rarely used sound that sounds like an RPG heal, this clears up any confusion as a result of it using the punch sound, which makes some less observant players ready to murder someone. The update makes the Chaplain just slightly more holy. Curse3 is the nicest sound I could find to represent healing (Suggested by Lynxems) while Blobattack sounds wet and squelchy, which is a far nicer sound to hear when you're whacking a lifeless body with the biggest scripture that can be found on the station!

Alternatives to Blobattack would be nice if found, but then, Blobattack seems to serve it's purpose of being a wet squishy sound.

This seems very minor, but I've seen this come up a few times personally.

### DIFF
--- a/code/game/objects/items/storage/book.dm
+++ b/code/game/objects/items/storage/book.dm
@@ -81,7 +81,7 @@ GLOBAL_LIST_INIT(bibleitemstates, list("bible", "koran", "scrapbook", "bible",  
 				H.update_damage_overlays()
 		H.visible_message("<span class='notice'>[user] heals [H] with the power of [deity_name]!</span>")
 		to_chat(H, "<span class='boldnotice'>May the power of [deity_name] compel you to be healed!</span>")
-		playsound(src.loc, "punch", 25, 1, -1)
+		playsound(src.loc, "curse3", 25, 1, -1)
 	return 1
 
 /obj/item/storage/book/bible/attack(mob/living/M, mob/living/carbon/human/user, heal_mode = TRUE)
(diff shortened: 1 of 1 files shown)


Affurit:
I’ll fix the name of this when I get in, probably. Didn’t notice that it was called Update Book.dm until I checked on my phone. I had to do this three times to avoid doing this from the Master branch because I’m awful with Github and inevitably ended up forgetting to name this thing.

code/game/objects/items/storage/book.dm
```diff
@@ -126,12 +126,12 @@ GLOBAL_LIST_INIT(bibleitemstates, list("bible", "koran", "scrapbook", "bible",
 		if(smack)
 			M.visible_message("<span class='danger'>[user] beats [M] over the head with [src]!</span>", \
 					"<span class='userdanger'>[user] beats [M] over the head with [src]!</span>")
-			playsound(src.loc, "punch", 25, 1, -1)
+			playsound(src.loc, "curse3", 25, 1, -1)
 			add_logs(user, M, "attacked", src)
 
 	else
 		M.visible_message("<span class='danger'>[user] smacks [M]'s lifeless corpse with [src].</span>")
-		playsound(src.loc, "punch", 25, 1, -1)
```
FlattestGuitar: I think this should stay here. When yous mack a corpse, you should just make the regular punching sound.

Affurit:
Fair enough ^.^

Affurit:
I completely forgot to remove the blob sound, it's done now!

FlattestGuitar:
No it's not.

Affurit:
Uh... I'm 100% sure I made that commit.

SmokingKilz:
If we can check through this I also think this is good to be merged.

PR #953, opened 2018-06-07
New Gas Sprites

Xantholne:
New/ported Sprites for Freon, Plasma, and N2O making them more "gas" and less "dots"
Plasma and N2O sprites are originally from a place called "Apollo" honestly I have no idea if that server even exists anymore.

:cl: Xantholne
add: New sprites for Freon, Plasma, and N2O gases
/:cl:

This is what I currently have.
gas

Current issues I discovered while doing this was:
Any visible gas (freon, plasma, even water vapor) if the sprite overlaps some object you want to interact with, it actually blocks you from using that object. 
However I am unsure what we'd need to do to fix this layering issue but it'll definitely be an issue.
Another thing was I was going to add tempature color changing fire, but could not successfully get that working and will need likely @FlattestGuitar help in working something out for that.


FlattestGuitar:
That issue will be present until we migrate to 512. Until then, this won't be merged, but afaik we're going to be doing that soon-ish anyway.

Cactus2388:
how does this affect glitter grenades?

StarSmasher44:
Thanks for returning those sprites! Apollo will love them :)
(I literally had no idea these existed, they were from before my time, but you may use them if you'd like to. We may, or may not.)

Xantholne:
Glitter bombs will need their icon patch changed to the plasmaOLD and such icons.

> Thanks for returning those sprites! Apollo will love them :)
(I literally had no idea these existed, they were from before my time, but you may use them if you'd like to. We may, or may not.)

Freon one is new, slightly altered to be more gaseous, feel free to use.

.travis.yml
```diff
@@ -4,8 +4,8 @@ dist: trusty
 
 env:
     global:
-    - BYOND_MAJOR="511"
-    - BYOND_MINOR="1385"
+    - BYOND_MAJOR="512"
+    - BYOND_MINOR="1424"
```
SamHPurp: Drop these changes, as we'll be going higher than 1424, and do it via a different PR.

PR #954, opened 2018-06-07
Crusher Changes

Xantholne:
Changed base Kinetic Crusher damage to 25, was 20.
Changed Kinetic Crusher recharge to 10, was 15
Changed Crusher Trophy drop rate to 20, was 5

The increase to the base damage and charge speed on the crusher allow it to still stay very competitive when compared to its' counterparts the Adv. Plasma Cutter, and Kinetic Accelerator that pretty much every miner will always time over the Crusher currently.

The increase to Trophy drop rates change it from the very unlikely 1 in 20 mobs being killed for a single trophy, to 1 in 5 mobs on average. I have yet personally to ever see a crusher trophy be dropped and neither has anyone else from non trivial monsters. Trivial monsters such as bubblegum remain the same with requiring 60% of damage dealt to them to have been from a kinetic crusher.

:cl: Xantholne
balance: altered the kinetic crusher to deal more damage and drop items at a higher rate. Buffs for everyone!
/:cl:

### DIFF
--- a/code/modules/mining/equipment/kinetic_crusher.dm
+++ b/code/modules/mining/equipment/kinetic_crusher.dm
@@ -8,7 +8,7 @@
 	name = "proto-kinetic crusher"
 	desc = "An early design of the proto-kinetic accelerator, it is little more than an combination of various mining tools cobbled together, forming a high-tech club. \
 	While it is an effective mining tool, it did little to aid any but the most skilled and/or suicidal miners against local fauna."
-	force = 20 //As much as a bone spear, but this is significantly more annoying to carry around due to requiring the use of both hands at all times
+	force = 25 //As much as a bone spear, but this is significantly more annoying to carry around due to requiring the use of both hands at all times
 	w_class = WEIGHT_CLASS_BULKY
 	slot_flags = SLOT_BACK
 	force_unwielded = 25 //It's never not wielded so these are the same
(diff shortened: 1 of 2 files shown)


FlattestGuitar:
With how hilariously underused the crusher is, this might just be the thing. I support this.

If people start overusing those, we can always waddle back.

Yamahawk:
After a bit of testing, it appears Crusher damage is currently 20 base, 70 detonating a mark, and 100 with a mark backstab.  Changing the base force from 20 to 25 effectively increases all the damage by 5 to 25/75/105.  This is excellent against targets you can't mark, like other players, but does very little to the mining mobs it's actually meant to be used against.

I'd suggest fiddling with the numbers in lines 104-105 and 109-110 to up the crusher mark damage and then adjust the description in lines 35-36 to match the new damage correctly.

Also need to change the force in line 11 from 20 to 25 for it to actually do anything to the base damage.

In regards to the drop rate, the comment still says 1/20 when it's been changed to 1/5, not a huge issue but may as well make it correct.

Ralta:
Perhaps a flat 25% increase to the three damage types would work, @yamahawk?

25 base, 93 marked (or 95 if you wanna make it more round, which makes sense for damage tracking), 125 backstab. Or even up the marked damage to 100? Probs need to test these numbers a bit.

Yamahawk:
Xanth corrected the damage changes to now be 25/95/135.

What this translates to in actual terms is:

3/1/1 hits to kill legions rather than 4/2/1
8/3/2 hits to kill watchers rather than 10/3/2
12/4/3 hits to kill goliaths rather than 15/4/3

So you can see there isn't really much of a difference between number of hits.  Until you factor in the damage increasing mods, which is just barely enough to push the damage up to 150 damage backstabs with a Bubblegum and Drake trophy, meaning you can now 2-shot goliaths with backstabs.  Having the initial damages just under the thresholds for less hits makes the trophies actually meaningful, so this seems like a good place to have them.

PR #956, opened 2018-06-07
Clockwork Spear + (Un)Holy Fire Weapons

Affurit:
This will reduce the pitchfork's brute damage by 8 but now the weapon deals 5 burning damage as a trade-off. I believe that the trade is worth it, but would not be averse to increasing the fire damage to 8 if it turns out that the null rod becomes too weak as a result. This seems unlikely to me. The Pitchfork will no longer fit in the bag of the Chaplain, or any thieves.  

Also adds a reskin of the Pitchfork, a Clockwork Spear, which has it's own sound effects and list of attack verbs. Clockwork cultists will see this weapon and the base null rod differently to normal crew now, having a clock-cult description of it instead of the normal crew description. I may do this for nar’sie cultists in the future.

The Clockwork Spear was asked for by a few people, who are always thirsty for Chaplain expansions, and also means that Narnar isn’t the only eldritch god getting some lovin from the Chaplain.

:cl: 
add: Added Burn Damage potential to Null Rod reskins.
add: Added Clockwork Spear, a holy weapon of Ratvarian design that burns the non-believers.
fix: The Pitchfork can now use the back slot and can no longer be put in the bag.
tweak: The Pitchfork and Clockwork spear trade off some damage for their burning ability.
/:cl:

### DIFF
--- a/code/game/objects/items/holy_weapons.dm
+++ b/code/game/objects/items/holy_weapons.dm
@@ -1,6 +1,6 @@
 /obj/item/nullrod
 	name = "null rod"
-	desc = "A rod of pure obsidian, its very presence disrupts and dampens the powers of Nar-Sie's followers."
+	desc = "A rod of pure obsidian, its very presence disrupts and dampens the powers of illicit magic."
 	icon_state = "nullrod"
 	item_state = "nullrod"
 	lefthand_file = 'icons/mob/inhands/weapons/melee_lefthand.dmi'
@@ -11,6 +11,8 @@
 	throwforce = 10
 	w_class = WEIGHT_CLASS_TINY
 	var/reskinned = FALSE
+	var/clockwork_desc = "A rod of Obsidian, a material known for corroding the Justiciar's creations."
+	var/bonus_burn = 0
 
 /obj/item/nullrod/attack_self(mob/user)
 	if(user.mind && (user.mind.isholy) && !reskinned)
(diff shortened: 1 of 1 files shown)


code/game/objects/items/holy_weapons.dm
```diff
@@ -11,6 +11,8 @@
 	throwforce = 10
 	w_class = WEIGHT_CLASS_TINY
 	var/reskinned = FALSE
+	var/clockwork_desc = "A rod of Obsidian, a material known for corroding the Justiciar's creations."
```
FlattestGuitar: Does this actually do anything? You just added a variable there, but didn't do anything with it.

code/game/objects/items/holy_weapons.dm
```diff
@@ -11,6 +11,8 @@
...
+	var/bonus_burn = 0
```
FlattestGuitar: Same here. You don't use that variable anywhere.

code/game/objects/items/holy_weapons.dm
```diff
@@ -390,8 +392,29 @@
 	lefthand_file = 'icons/mob/inhands/weapons/polearms_lefthand.dmi'
 	righthand_file = 'icons/mob/inhands/weapons/polearms_righthand.dmi'
 	name = "unholy pitchfork"
-	w_class = WEIGHT_CLASS_NORMAL
+	w_class = WEIGHT_CLASS_BULKY
 	desc = "Holding this makes you look absolutely devilish."
 	attack_verb = list("poked", "impaled", "pierced", "jabbed")
+	slot_flags = SLOT_BACK
```
FlattestGuitar: You add a new slot to carry the thing, but you don't add a sprite for doing so. Is there one already?

code/game/objects/items/holy_weapons.dm
```diff
@@ -390,8 +392,29 @@
...
+	righthand_file = 'icons/mob/inhands/weapons/clockwork_righthand.dmi'
+	name = "clockwork spear"
+	desc = "A pointy spear made of holy brass. It ticks and tocks."
+	attack_verb = list("ticked", "tocked", "singed", "speared")
+	clockwork_desc = "A disgusting and cheap replica created by the church to imitate the divine armory of the Justiciar."
+	hitsound = 'sound/magic/clockwork/ratvar_attack.ogg'
+	
+/obj/item/nullrod/attack(mob/living/target, mob/living/carbon/human/user)
+	. = ..()
+	if(!QDELETED(target) && target.stat != DEAD)
```
FlattestGuitar: This entire proc can be done with

```
if(target.stat != DEAD)
    target.adjustFireLoss(bonus_burn)
```

code/game/objects/items/holy_weapons.dm
Affurit: There’s no sprite for another back slot on the holy weapons list unless I’m just dumb and blind, and while I don’t agree with just piling up undone work on someone (myself), this solves the issue of having nowhere to carry it yet. There’s a precedent for items that are invisible in slots, particularly the belt for the children of claymores. I’ll make this a priority if possible?

code/game/objects/items/holy_weapons.dm
Affurit: This allows us to modify how much burning a particular re-skin does in the future if we need to do so ^.^

I hope my responses don’t come off as defensive, I’ve taken into account your point about the slots, and am fully open to criticism.

Affurit:
Perhaps I’ll cut out the Clockwork spear. It was a nice, quick, easy job to reskin the pitchfork. We already have a cult blade and people seemed interested in this was all I was thinking.

The burn damage can also be undone but I do rather like all of the holy weapons being different and unique. 

The new slot and weight of the pitchfork need to stay though. 

There are a number of null rod reskins that are purely visuals, I don’t really see the harm in this in particular, but I can see the point about overloading people’s choices.

SmokingKilz:
Variety =/= Watering down. 

That being said, I would like to see the general reaction from the community first.

Affurit:
Closing and re-doing since there's some stuff I missed that is on my test build.

PR #957, opened 2018-06-07
New Medical Hud Icons

Xantholne:
Adds new Medical Hud Icons condensing the giant 3x32 gross bar into a fancy little heart based off of and inspired by Goon's health heart system.

At 100% you will have a lovely green Heart that slowly changes from Green, Yellow, Orange, Red, Red/White, Red/Black, and finally Black for death.

Because of it's position I made a new Good/Bad virus icons based off of an actual cell and placed it on the right side to avoid covering the individual what so ever.
To keep it still easier to identify, the Xeno Embyro icon was moved over to the right side and left untouched. It overlays ontop of the viral icons while still letting them have some strands stick out for keener eyes to notice.

There are several benefits to using this new HUD system over the current one, and only one or two cons that I personally can think of.

Pros:
It's much more clean, and sleek.
It provides an RPG element
It doesn't stretch across the entire tile the person is on
It seems more logical than a giant bar that covers a 3x32 area on the top of the tile the person is on
It's very unique

Cons:
People actually have to put in a minor amount of effort to realize someone is injured.

:cl: Xantholne
add: Medical HUDs have a new layout for a person's health based on a heart design going from healthy Green, to Yellow, to Red, to crit flashing Red/White, and hard crit flashing Black/White before eventual death, Black. Viruses also have a new icon on the opposite of the new heart representing if they have a beneficial virus, or harmful virus!
/:cl:

med hud preview 2
-


SmokingKilz:
I mean as I stated on discord last night I like it, while I agree it does look a little "gamey" for what is meant to be IC health HUD's, I still with it looks neater than the bar system we've currently got.

Cactus2388:
I don't see a problem with our current health HUD and frankly think it looks better.

DaveTheHeadcrab:
Looks great, infinitely better than the uggo bar.

FlattestGuitar:
The virus has a completely different style to the heart. Either keep the outlines or not.

This is far too inconsistent with itself and the rest of the HUDs.

Xantholne:
Updated virus icons to share the similar dark edges and still have a bright center to see the icon

They're also now a miniature biohazard symbol

Ralta:
I love it, but I honestly don't think the heart is good enough. The bar is fully graduated, meaning you can see if someone is declining in health immediately. This would require them dropping a significant amount before it became readily apparent (unless you spam them with an analyser).

Xantholne:
The issue there is. Sure you can see they have a whole whopping 10 DMG more easily with the bar than the heart. They share the same color pallet except the heart has even more. That bar is 3 pixels that stretch across the entire tile and are only accurate barely until they've taken 80 DMG. That's needless to say how horrifying they actually look. With the current bar design even you cannot see how hurt they are at a glance by the time they entire the negative range. The current negative range for the health bar is a very hard to see "dark bar" that starts covering the first frame. But every negative health Sprite has 2 frames. And that's not all. The current bar literally flashes a giant Ted overlay on top of the more important information. There's no color change. Thetrs nothing that changes the difference between -10 and -40. By the time you can finally tell the difference thanks to the giant cluster fuck bar flashing over it is by the time they reach -70. By then they're moments from death. And even then it is neigh impossible to really gage their health until they are at -80 -90 were it's just an epilepsy attack of flashing red. None of this is the case with the heart. The heart had varying colors even in negative health making it more clear how injured someone is compared to others. It won't even trick you into thinking someone is close to death until they literally care a minute away. 
The onlu cons I can see from it are they people have to deal with change.

Ralta:
Rather than make the assumption that people can't live with the change (as opposed to constructive criticism), perhaps you should consider how how this works in-game.

If you honestly believe that a bar is hard to read at a range, you probably don't ever use the medical HUD in-game. At a glance, I can see someone's current health, and if they're declining (as well as how quickly). It's far from perfect and could certainly be improved on but I think it works better than basic colour changes.

I'd be happy with the heart if it perhaps starting getting less full as you take damage, akin to the bar. Like if you had an outline of the heart, and as you take damage, the level in the heart goes down. Or something like that, anyway. Could still keep the colours of course.

SmokingKilz:
@Ralta Brings up some valid points, I feel as though a fair middle ground would to make it so the heart started to slowly get less full as suggested?

Xantholne:
Been trying to refine the new walls I've been meaning to PR too much, completely forgot I was going to experiment with making the border less bold and a much more drastic color change towards green in the first 40 health.

SamHPurp:
Please commit the removal of the DME

FlattestGuitar:
Honestly, if we want to go in the artsy direction over clarity here, the way Bay does medhuds is way better.

 You get one of those ekg graphs above the spaceman's head and they look swood as heck

SamHPurp:
Author said they were going to close it, so I assume this is dead. Let me know if not.

PR #958, opened 2018-06-07
Clockwork Spear + (Un)Holy fire V2 

Affurit:
Same as the last one, not sure if it's possible to edit stuff. 

Fixes some things that were on my test build but were never updated into the pull request!

Added the Clockwork effect to the Clockwork Spear, which means it can now benefit from the Clockwork_desc text.

Sorry if this is the wrong way to do things, Github is weird to me.

:cl:
add: Added the Clockwork Spear to the list of holy weapons.
fix: Changed the weight of the Pitchfork holy item.
tweak: Clockwork Spear and Pitchfork sacrifice some brute damage for burn damage.
/:cl:


code/game/objects/items/holy_weapons.dm
```diff
@@ -390,8 +391,31 @@
...
 	attack_verb = list("poked", "impaled", "pierced", "jabbed")
+	slot_flags = SLOT_BACK
 	hitsound = 'sound/weapons/bladeslice.ogg'
 	sharpness = IS_SHARP
+	force = 10
+	bonus_burn = 5
+	
+/obj/item/nullrod/pitchfork/clockworkspear/clockwork 
+	icon = 'icons/obj/clockwork_objects.dmi'
+	icon_state = 'ratvarian_spear'
```
FlattestGuitar: Using single quotes like this indicated you want BYOND to find a file that matches, not a string. Use `" "` instead.

code/game/objects/items/holy_weapons.dm
```diff
@@ -390,8 +391,31 @@
...
+	item_state = 'ratvarian_spear'
```
FlattestGuitar: Same here.

Affurit: I managed to PR an alternative version (again) rather than the actual updated version that worked perfectly fine in all the testing I did on it...

code/game/objects/items/holy_weapons.dm
```diff
@@ -390,8 +391,31 @@
...
+/obj/item/nullrod/pitchfork/clockworkspear/clockwork 
+	icon = 'icons/obj/clockwork_objects.dmi'
+	icon_state = "ratvarian_spear"
+	item_state = "ratvarian_spear"
+	lefthand_file = 'icons/mob/inhands/antag/clockwork_lefthand.dmi'
+	righthand_file = 'icons/mob/inhands/antag/clockwork_righthand.dmi'
+	name = "clockwork spear"
+	desc = "A pointy spear made of holy brass. It ticks and tocks."
+	attack_verb = list("ticked", "tocked", "singed", "speared")
+	var/clockwork_desc = "A disgusting and cheap replica created by the church to imitate the divine armory of the Justiciar."
```
FlattestGuitar: This variable does absolutely nothing. Either remove it or use it. You could do something like this:

```
/obj/item/nullrod/pitchfork/clockworkspear/clockwork(mob/user)
. = ..()
if(*some check to see if user is a clockcultist*)
    to_chat(user, clockwork_desc)
```

Not sure about the exacts there, but look around for a different `examine()` proc implementation and you should see what you should do here.

code/game/objects/items/holy_weapons.dm
```diff
@@ -390,8 +391,31 @@
...
 	name = "unholy pitchfork"
-	w_class = WEIGHT_CLASS_NORMAL
+	w_class = WEIGHT_CLASS_BULKY
 	desc = "Holding this makes you look absolutely devilish."
 	attack_verb = list("poked", "impaled", "pierced", "jabbed")
+	slot_flags = SLOT_BACK
 	hitsound = 'sound/weapons/bladeslice.ogg'
 	sharpness = IS_SHARP
+	force = 10
+	bonus_burn = 5
```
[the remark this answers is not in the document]
Affurit: The original weapon does burn damage, which is where I actually borrowed the code from.

code/game/objects/items/holy_weapons.dm
```diff
@@ -390,8 +391,31 @@
 	lefthand_file = 'icons/mob/inhands/weapons/polearms_lefthand.dmi'
 	righthand_file = 'icons/mob/inhands/weapons/polearms_righthand.dmi'
 	name = "unholy pitchfork"
-	w_class = WEIGHT_CLASS_NORMAL
+	w_class = WEIGHT_CLASS_BULKY
 	desc = "Holding this makes you look absolutely devilish."
 	attack_verb = list("poked", "impaled", "pierced", "jabbed")
+	slot_flags = SLOT_BACK
```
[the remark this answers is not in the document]
Affurit: I’ll sprite them when I get home

code/game/objects/items/holy_weapons.dm
Affurit: The spear should be on the list of clockwork items, which means that Ratvarian cultists would get an alternate examine.

Affurit:
Fixes most of the issues raised and fixes the carp’sie plushie.

SamHPurp:
Doing a mix of damages makes them stronger, because they become much harder to heal... not so sure about that part.

Also not sure about continuing the trend of giving the chaplain more and more weapons / memes.

Affurit:
Burn and brute damage on one weapon means that you have to apply two types of chems or treatments. There’s also the fact the fire damage does its damage to each separate limb which makes it annoying unless you have burn patches.

The Chaplain can only have 1 null rod at a time. Adding more weapons increases his options but not his base level of power.

SamHPurp:
Your first paragraph contradicts the second.

Increasing the amount of types of treatments/chems/locations makes it inherently stronger than one that only effects a single location with a single damage type.

Affurit:
Hence the damage nerf that came with it originally, which I believe I reverted on someone else's word since some other points were raised.

I refuted a point you made. This alone doesn't empower the Chaplain, it offers more choice. The Chaplain can still choose the Bo Staff which would deal more damage to one limb and grant a block chance. 

Chaplain weapons aren't toys, this is some weaksauce stuff still.

Affurit:
This has been re-done.

PR #960, opened 2018-06-07
METASTATION - Removes chaplain soul shard

Quantum-M:
Removes the Chaplain's soul shard from metastation.

:cl: 
del: Removed Chaplain soul shard from meta station
/:cl:

The chaplain soul shard tends to bring several issues, especially if the chaplain is not a antag. The times it has been used, the chaplain has resorted to soul shard people without good reason, and even then, a normal person, RP wise, doesn't have a reason to be killed and become a ghost.

It also encourages the chaplain to validhunt in my opinion, soul sharding any antags if they happen to come across any. It's also a instant body destroying item, once someone gets sharded, they are dead, unable to revive, and end up stuck as a crappy simple mob, and the Chaplain _already_ has the cremator available, it shouldn't be that easy to fully remove someone from the round.

Lastly, the soul shard is primarily seen as a cult item, it makes no sense for a chaplain to have a item that is used by a rival, violent religious cult.


SmokingKilz:
I mean there's the works of replacing it with something more holy-like. I'd like to see the current soul shard get replaced with that.

Quantum-M:
That would be a alternative to this then, @SmokingKilz .

Ziiro:
+1 x 1000. I hate that thing.

Cactus2388:
yes
that is all

AdamElTablawy:
RIP the one counter to lich wizards

Ralta:
Remove it for now. If someone had a better idea, by all means add it. But this thing is awful.
+1

Quantum-M:
@FlattestGuitar , I think only metastation is the only map with the shard. And if someone is doing a rework for it, they can simply re-add it.

SmokingKilz:
Yeah the rework is no longer just a simple switch, this can be merged. 


@AdamElTablawy Lich's can still be gibbed/cremated which is one of the most popular way people dispose of them anyway.

PR #972, opened 2018-06-09
Adds Tons of New Cyborg Models

Xantholne:
Cyborgs of all make and models have been added to literally every single module selection.

After selecting a module you can now choose one singular model to base yourself off of!

This is a one time pick, so pick by name carefully or else you'll have to spend a minute if the roboticist is competent enough to have them reset your module.

:cl: Ported by/Made by Xantholne and made by various Paradise spriters
add: Borgs have tons of new models they can pick after module selection
/:cl:

borg preview
WOW JUST LOOK AT ALL OF THIS GORGEOUS TRASH, I MEAN BORGS.
THAT I HAD TO SPRITE NEW EYES AND PANELS FOR.
EVERY.
SINGLE.
ONE.


Ralta:
Yiss, it has Trav's sprites!

SamHPurp:
It is my opinion that having multiple forms of a singular entity all doing the same thing - is bad game design.

I'd rather a singular model for each of the roles.

So if people wanted to select amongst these to replace existing, rather than extend an option.... but each type should have a single coherent visualisation.

SmokingKilz:
While I am usually all for extending choice for things, I do agree with Purp on this. I think we should all choose one of these for each type and make it so only one sprite per borg type exists.

Xantholne:
The issue with that for the, cricket, noble, and standard, is that everyone prefers one over another in that set for X module.

Someone could prefer let's say Noble Engineer over any other Noble. They find it looks best.
But someone else could prefer Noble Service over any other Noble models.

So how do we prune them then to only allow X model for Y module if people have varying tastes?

FlattestGuitar:
>So how do we prune them then to only allow X model for Y module if people have varying tastes?

We close this PR?

I'm sorry, but this is something that we've been reiterating over the past 6 months. We don't want more cyborg models. Cyborgs need to clearly communicate their function and they're the last element of gameplay that should be subject to this level of customization. 

A separate issue is that the styles of the new cyborg sprites are horribly inconsistent. Some of the models have really striking shading, while others  have barely anything. The NOBLE borgs are gigantic and don't look like the other cyborgs at all.

No from me.

SamHPurp:
`So how do we prune them then to only allow X model for Y module if people have varying tastes?`

You don't, you either make a case for the ones chosen, and we roll with it... Or leave them as is.

This topic has come up a lot over the last 6 months or so, and this has consistently been my response.

Also on the topic of the NOBLE borgs, they're still too mech-like for me to be happy with.

PR #978, opened 2018-06-11
METASTATION - Cafeteria addition

Quantum-M:
:cl: optional name here
tweak: Expanded the kitchen
tweak: Moved bar into the club area, and made the theater area smaller.
tweak: Art storage removed, its items moved to dorms.
tweak: Cafeteria added to the spot where the bar and art storage was previously.
/:cl:

Map changes to metastation. The intent is to have a common area for the crew to be able to go and eat together, along with giving a more general area to the bar itself. The bar was moved into the club, along with its storage moved behind it into maintenance. This led to the theater to be made way smaller to accommodate more club stuff.

Art storage was given the finger, deleted, and its items moved into dorms, in the fitness room to be specific.

The left over area was given for a cafeteria. Its first entrance is the same as previous, but a northern entrance was given.

The kitchen was also expanded one tile to the north, to make it less claustrophobic and give the chef more space.

EDIT: I almost forgot. The kitchen was given anti-greytide measures, namely in the form of windoors at the counters.

bar

cafeteria

art

I did a small change a while ago but forgot to mention it. The bar has another disposal bin for customers and whatnot, and some of the stuff in the bar maintanance was moved so the mulebot entrance for the bar backroom wasn't blocked. Oh and some extra holopads for the AI in the cafeteria and bar.


DaveTheHeadcrab:
I like it. Good usage of the casino area that was fairly empty before.

LYNXEMS:
Pretty good, just think anti greytide measures are not needed, because if there is no chef someone still has to cook...

FlattestGuitar:
Not mapmerged. Quantum pls

AffectedArc07:
Flattest, mapmerge is broken

scrubmcnoob:
This is great.

PR #979, opened 2018-06-13
Removes QM Access from irrelevant jobs

Quantum-M:
:cl: 
tweak: Removes QM access from Shaft Miner and Cargo Technician.
/:cl:

Removes the QM access that low pop Shaft Miners and Cargo Technicians has. The access is meant for the one who oversees cargo and mining, does not help in anyway cargo or mining, and, most of the times, the stuff behind the access ends up being looted by Cargo Technicians, especially the cape and the megaphone.


FlattestGuitar:
This is an attempt to disempower the proletariat.

:100: x :-1:

PR #981, opened 2018-06-13
Added a cool-down for Resting

Bksparkie1:
:cl: Bksparkie1
tweak: Resting now has a much longer cooldown.
/:cl:

[why]: # Stops people from spamming rest to avoid getting shot (breaks disco machine kinda... working on a fix)

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -291,11 +291,11 @@
 
 /mob/proc/get_contents()
 
-/mob/living/proc/lay_down()
+/mob/living/proc/lay_down(var/force = FALSE)//force will override the cooldown
 	set name = "Rest"
 	set category = "IC"
-	
-	if(resting_cooldown > world.time)
+
+	if(resting_cooldown > world.time && !force)
 		return
 
 	resting_cooldown = world.time + resting_cooldown_duration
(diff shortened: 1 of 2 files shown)


FlattestGuitar:
Way too many variables. Just use a single one where you save the time of the last use. All checks can be extrapolated from that.

I'd also cut the spam that says it's too early. It's pretty pointless with a cool down this short.

code/modules/mob/mob_defines.dm
```diff
@@ -49,8 +49,11 @@
 	var/spacewalk = FALSE
 	var/druggy = 0			//Carbon
 	var/confused = 0		//Carbon
-	var/resting = 0			//Carbon
-	var/lying = 0
+	var/resting = 0
+	var/resting_cooldown_duration = 10 // Roughly 5 seconds
```
GuppyLaxx: you mean 1 second

code/modules/mob/living/living.dm
```diff
@@ -291,10 +291,14 @@
 
 /mob/proc/get_contents()
 
-/mob/living/proc/lay_down()
+/mob/living/proc/lay_down(var/force = FALSE)//force will override the cooldown
```
FlattestGuitar: Weirdly enough, disco balls already use the parameter, even though it's never defined. I'd imagine that's exactly what it was for.

PR #992, opened 2018-06-18
Bad href exploit, bad! *whacks with newspaper*

FlattestGuitar:
*cough cough*


Ralta:
Your comment should definitely say "//NOW we're checking against this list to prevent href shenanigans."

PR #996, opened 2018-06-19
Death to Cat Ears, part Two

Quantum-M:
:cl: 
del: Removes left over references of cat ears from the preferences files.
/:cl:

Guts out what is left of the cat ears nonesense that wasn't found before hand.


Iamgoofball:
doing God's work

PR #997, opened 2018-06-19
Fixes issues with stack crafting

### LINKED ISSUE #833: Crafting an item stack on top of another only adds 1 item to the resulting stack (and the biogenerator, again)
I've been experimenting with item stacks interactions and I've tried to compile a list of all the stuff I've found so far. There are a few inconsistencies, and there are also some clear bugs, although I was expecting more issues with vendors/machines. Here are all my findings.

A) With an item stack on the floor:
* dropping a stack from your hand into it - combines correctly
* dragging another stack into it - combines correctly
* throwing a stack from afar - _they don't combine. Not sure if this is intended?_
* **crafting a stack on top of another (e.g. Tiles) - doesn't combine properly, it only adds 1 item to the stack regardless of crafted amount. More examples below.**

B) Specific instances of broken crafting (the ones I've found):
* Tiles: steel, gold, silver, wooden...
* Steel rods
* Sandstone - If you try to make sandstone out of a bunch of sand and there's already a sandstone stack in the tile, it'll add 1 item to the stack and consume all the sand
* Grass/Carpet tiles using the grass/carpet plant - same as the sandstone, it'll consume all the plants in your tile but it'll only add 1 item, if the tile already had a stack in it.
* Probably any other crafting where the result is a stack of items

C) Vendors and machinery
* **The botany biogenerator still doesn't properly stack leather or cloth.**
* Machines that worked correctly when tested: ORM, Autolathe(metal, glass, cable coil, etc), Protolathe/Circuit imprinter/Exosuit fab (ejecting materials from storage), NanoMed Plus (medical gauzes), Mining vendor (mining beacons). I stopped testing after those.

------
So yeah, crafting is a bit broken. And the biogenerator, still.

**Steps to reproduce:**

1. Grab a bunch of metal or any other material that can be made into tiles (or any other item that's produced in stacks).
2. Grab something in your offhand so the resulting crafted item falls on the ground.
3. Craft a stack. If it's tiles, there'll be a 4 item stack at your feet.
4. Craft another stack. Now there'll be 5 items in the stack, instead of 8.

Alternatively,

1. Harvest/admincreate a bunch of grass in one spot (works with sand/ash to produce sandstone as well).
2. Grab 1 grass, move away and use it to create tiles out of it (probably makes 2).
3. Move back to the position with several grasses in it, drop the grasstile stack in it.
4. Grab and use another grass. This will consume all the grasses at your position, but only add 1 item to the stack of grasstiles instead of the correct amount.

Sikroth:
My first PR, yay. This is for the most part a port of https://github.com/tgstation/tgstation/pull/37760 with some minor additions. I also ported some changes from the stacks.dm file (couldn't find the exact PR they're from) because there were actually two separate issues with stack crafting.

:cl:
fix: Fixes several issues with crafting stacks of items that resulted in only 1 item being created.
/:cl:

This PR hopefully fixes #833 , and I'm also hoping it won't introduce any new issues. All the tests I did seemed to work okay.

### DIFF
--- a/code/game/mecha/equipment/tools/work_tools.dm
+++ b/code/game/mecha/equipment/tools/work_tools.dm
@@ -302,8 +302,7 @@
 	var/max_cable = 1000
 
 /obj/item/mecha_parts/mecha_equipment/cable_layer/New()
-	cable = new(src)
-	cable.amount = 0
+	cable = new(src, 0)
 	..()
 
 /obj/item/mecha_parts/mecha_equipment/cable_layer/can_attach(obj/mecha/working/M)
@@ -360,8 +359,7 @@
 			m = min(m, cable.amount)
 			if(m)
 				use_cable(m)
-				var/obj/item/stack/cable_coil/CC = new (get_turf(chassis))
-				CC.amount = m
+				new /obj/item/stack/cable_coil(get_turf(chassis), m)
 		else
 			occupant_message("There's no more cable on the reel.")
 	return
--- a/code/game/mecha/equipment/tools/work_tools.dm
+++ b/code/game/mecha/equipment/tools/work_tools.dm
@@ -334,8 +334,7 @@
 		if(to_load)
 			to_load = min(target.amount, to_load)
 			if(!cable)
-				cable = new(src)
-				cable.amount = 0
+				cable = new(src, 0)
 			cable.amount += to_load
 			target.use(to_load)
 			occupant_message("<span class='notice'>[to_load] meters of cable successfully loaded.</span>")
--- a/code/modules/hydroponics/grown/grass_carpet.dm
+++ b/code/modules/hydroponics/grown/grass_carpet.dm
@@ -36,17 +36,7 @@
 			continue
 		grassAmt += 1 + round(G.seed.potency * tile_coefficient)
 		qdel(G)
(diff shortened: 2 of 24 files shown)


code/game/mecha/equipment/tools/work_tools.dm
```diff
@@ -302,8 +302,7 @@
 	var/max_cable = 1000
 
 /obj/item/mecha_parts/mecha_equipment/cable_layer/New()
-	cable = new(src)
-	cable.amount = 0
+	cable = new(src, 0)
```
SamHPurp: Lines 337 and 338 need this treatment as well, right?

code/modules/hydroponics/grown/grass_carpet.dm
```diff
@@ -36,17 +36,7 @@
...
-		GT = new stacktype(user.loc)
-	GT.amount = grassAmt
-	for(var/obj/item/stack/tile/T in user.loc)
-		if((T.type == stacktype) && (T.amount < T.max_amount))
-			GT.merge(T)
-			if(GT.amount <= 0)
-				break
+	new stacktype(user.drop_location(), grassAmt)
 	qdel(src)
 	return
```
SamHPurp: You don't need a trailing return here, might as well remove it as you fix the other thing.

PR #1002, opened 2018-06-23
Crisper Fonts For Status Displays & Brig Timers

AndrewMontagne:
image

Changed the font to a default raster font present on all modern Windows systems, result is a crisper text that's more readable.

:cl: AndrewMontagne
tweak: Tweaked status display font and maptext alignment.
/:cl:


alexkar598:
Yay!the text is finally readable

FlattestGuitar:
It says `abcde fghij`. I think that leaves a bit to be desired, since I literally can't read half those letters.


Here's one with size upped to 6 and one y higher than what this PR has. It's definitely readable, but with capitals is goes off the screen bit of the screen. I'd rather not lower the letter count to 4 per line.

The way I see it, it's going to be one or the other here. Possibly forcing all caps and using 5pt? What do the people think?

SamHPurp:
You first image isn't easily readable.... but your second one literally overlaps on pixels from the monitor you shouldn't be able to draw onto.

FlattestGuitar:
That's literally what I said! The first one is what the PR is eight now. Perhaps we could keep that letter size bit force caps on what the screen displays?

AndrewMontagne:
Before:
image

After:
image

PR #1010, opened 2018-06-28
Adds a map to the PDA

AffectedArc07:
Added a way to view the map of the current station, right from your PDA. It will automatically pull the correct map, depending on the current station loaded. 

Main Menu:
image

image

Possible issues with acceptance of this PR:
- Webmap is not hosted by Oracle itself

:cl: AffectedArc07
add: You can now view the current station map on your PDA
/:cl:

Why: People ask what is this map, also useful for engineers to see wiring grid.


FlattestGuitar:
We will need to either host the webmap ourselves or use the preexisting crew monitor thingy.

The webmap is way prettier, but it requires updating by hand, while the crew monitor map is uggo, but is generated on its own.

What do you think, random people on our GitHub?

Quantum-M:
Crew monitor would probably be easier, but it's rather laggy at times when you open it, no idea how a webmap would be like performance wise.

FlattestGuitar:
image
Use this pic tho.

AffectedArc07:
New icon:
image

Cactus2388:
The crew monitor map is pretty laggy not sure how laggy this would be either.

ShattForte:
I can see this being extremely helpful to engineers, as you can see the pipe/wire work. Makes repairs a breeze, I already keep DM open with the current map but if this was in the game it'd bypass the extra work.

SamHPurp:
This seems like it'd be lagging, put extra strain on things, and links to non-server owned space.

👎

FlattestGuitar:
Well, we could host the webmap ourselves...

Still, odd that we wouldn't use the already existing map rendered for the crew monitor.

Not sure what to do here, to be honest. 🤔

AffectedArc07:
We can use the existing map, but you can already guess how shit the quality will be

AffectedArc07:
https://github.com/OracleStation/OracleStation.github.io/pull/6

SamHPurp:
This isn't something we want, after many discussions.

Do come talk to me and we can cover all the reasons again. Sorry AA

PR #1029, opened 2018-07-03
Blacklists another build

SamHPurp:
See: http://www.byond.com/forum/?post=2378537

Also rolls us back off of that build. Rip.


AndrewMontagne:
Current version is 512.1434. Do we just want to upgrade to that instead?

PR #1038, opened 2018-07-07
Fobbing off AHelps to MHelps

AndrewMontagne:
image

No changelog as it's admin-only. Added an extra option for admins to defer tickets to mentorhelps.


FlattestGuitar:
I personally prefer the people being able to pick from more options, along the lines of "Ask a question", "Report a player", "Report an issue", "ask for antag", instead of bloating the command bar further. It's already hard enough to find the right button.

AndrewMontagne:
@FlattestGuitar That's more along the lines of redesigning the front end of the help system entirely, which is a much bigger chunk of work. This is a simple solution to an issue we currently have.

SmokingKilz:
While that command bar is already large it's really not hard to find what you need. @FlattestGuitar makes some good suggestions, but I think what Andrew's put together is a really great tool which helps administrators focus on administrating when the server is busy and still makes it so questions can be answered.

AndrewMontagne:
Selfmerging this. Sue me.

PR #1040, opened 2018-07-08
Sec jacket

FlattestGuitar:
The first one is the Sec jacket, the last two are the Desert and Navy ones.

The Sec jacket has worse armor than the Security Vest, but will also protect your hands in a fight. I really don't want to write a script that will simulate a fight to check whether they're fair, but I think the bigger risk is the jacket being underpowered, so whatever. I doubt this will matter anyway.

:cl: Flatty
add: Navy and Desert military jackets are now in the Autodrobe
add: A new, swanky, beautiful, smashing Security Jacket will now sometimes spawn in a Security locker. Grab it before someone else does.
/:cl:


ShattForte:
Looks great.

Ralta:
Nice

PR #1041, opened 2018-07-08
Corporate Uniform

FlattestGuitar:
**RED-SLEEVED VERSION NOW**

Stats the same as a regular uniform. If we ever do a loadout system this will be one of the choices.

Y/N?

This PR also fixes an issue with the existing red uniform where the armband would be present on both the right- and left-facing sprites, even though it's only on one arm.

:cl: Flatty
add: Added a new uniform and cap for Security. Present in the red wardrobe inside of the Brig.
fix: Regular Security uniform now has a proper armband sprite
/:cl:


FlattestGuitar:
Added the digitigrade variant.

SmokingKilz:
I love it, well done.

PR #1042, opened 2018-07-08
Booze machines don't need any access anymore

FlattestGuitar:
:cl: Flatty
tweak: Most booze-o-mats no longer require any access on your ID.
/:cl:


Ziiro:
power creep

Quantum-M:
I kind of disagree. For booze machines in maintenance or ones created later in the round, it would be fine, but the one in the actual bar I think should have access restrictions, as it will encourage people (aka greytide) to go over the counter and loot the machine if the bartender steps out for a moment.

AdamElTablawy:
helps chemists make neurotoxin if they can just grab 5 bottles of the requisite and run instead of having to do back and forths for the booze dispenser, tbh

SamHPurp:
Disagree with this. Built ones, sure, have no restrictions... but the actual one in the bar? No. This'll only encourage people to ignore the barkeep and jump the bar.

PR #1047, opened 2018-07-10
ID Card Ground Icons

DaveTheHeadcrab:
:cl:
add: Add support for items that have different inventory/ground icons.
imagedel: Adds ground icons for our ID cards.
/:cl:

This PR adds a system that allows us to specify a different icon_state or icon file for items when they are in the player's inventory or a storage item and on the ground. ie; Use a "thumbnail" icon for an item while it is on the ground so it doesn't take up as much space visually, and a more detailed/larger version of the icon while it's in the inventory.

If an inventory state/icon file is not specified, it will default to the initial inventory state/icon file.

This system in action:
headcrabdance

This PR debuts this system by adding a second set of ID card icons that are smaller for use in the world. We keep our large and pretty(?) ID card icons for use in-inventory. This is in (late) response to a few complaints about the ID card icons being pretty but taking up too much room, so hopefully this system is an acceptable compromise!

Ground IDs:
groundids - copy

Inventory IDs:
inventoryids


Ralta:
Looking forward to seeing this spied to lotsa things!

PR #1057, opened 2018-07-12
HoP and Cap get those cute mini eguns instead of the big-ass ones

FlattestGuitar:
It was talked about earlier at some point. I like the idea. Discuss.

:cl: Flatty
tweak: The Head of Personnel and Captain will start with smaller e-guns for more portability, but less firepower.
/:cl:

To clarify:
* 40% less ammo
* in-built flashlight
* fits in pockets or up your ass
* the same as a regular e-gun in every other regard

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/security.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/security.dm
@@ -29,7 +29,7 @@
 	new /obj/item/clothing/gloves/color/captain(src)
 	new /obj/item/restraints/handcuffs/cable/zipties(src)
 	new /obj/item/storage/belt/sabre(src)
-	new /obj/item/gun/energy/e_gun(src)
+	new /obj/item/gun/energy/e_gun/mini(src)
 	new /obj/item/door_remote/captain(src)
 
 /obj/structure/closet/secure_closet/hop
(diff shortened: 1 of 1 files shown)


DaveTheHeadcrab:
What are the differences?

FlattestGuitar:
Good point.

* 40% less juice to shoot with, though each shot is just as damaging as on the big thing
* built-in flashlight
* small item instead of a normal one, so fits in a pocket or wherever a bored HoP migh want to put it

Battletrap360:
Sounds great.

Quantum-M:
I think the Head of Personnel should only get it, while the Captain keeps the old regular one.

My reasoning is that the Captain is a higher target for antags compared to the HoP (e.g. Automatic all access, nuke disk, hand tele, antique gun, space jetpack).

SmokingKilz:
I'm not sure about this, I've been a loud voice in wanting all our weapons re-sprited anyway, till then I'm not sure either way with this PR.

SamHPurp:
That sounds like a significant buff... 40% more shots, at the same damage... and you can hide them in smaller places....

I'm against this unequivocally.

FlattestGuitar:
@SamHPurp they have 40% less energy in the cell, meaning 40% fewer shots you can take. It's a nerf in that regard. I wasn't clear on that.

Ziiro:
Make them a traitor objective

Ralta:
I think this would be good for balance. He Captain and HoP should not be a force to be reckoned with, they should be a force that has the means to defined themselves a little.

The Captain has their sabre and the antique laser rifle etc anyway, plus the Blueshield following them around like a lost puppy.

My ONLY concern is Captain's rushing the armoury at roundstart to 'supplement' their equipment. 

Also, yes make these a traitor objective. Miniaturised weapons tech could be important for the Syndicate!

PR #820, opened 2018-04-05
Kill urself

FlattestGuitar:
:cl: Flatty
tweak: Double Agents will no longer get the objective to kill themselves. Relax at the bar instead.
tweak: Traitors will no longer get the objective to kill themselves; they get nothing instead. Still end your life if it floats your boat.
/:cl:

PR #1062, opened 2018-07-14
Removes the "die a glorious death" objective.

FlattestGuitar:
A reopen of a part of #820.

This PR removes the `die a glorious death` objective from regular traitors.

This PR doesn't remove the interactions possible right now. You will now have one free objective slot instead of that old one, which means you yourself can choose to get yourself killed or escape on the shuttle. The "free slot" thingy has the same chance and prerequisites for being picked as the old objective.

Here's my reasoning:
* The objective makes it seem like you are allowed to do anything "glorious" (i.e. acts of terrorism) to complete it. That is wrong, and doing so is a breach of rules that might lead to a ban, as it already has, multiple times.
* Some people might not want to die and forcing them to choose between greentext and having their character survive the round is odd.
* People who want to powergame the objective to win will just silently kill themselves in maint, which is really uneventful.
* The only reasonable way to accomplish this objective in a manner that is interesting to you and other players is to specifically let Security kill you during a chase, which is so rare that we might as well let people choose whether to do it themselves.

:cl: Flatty
tweak: Removes the "Die a glorious death" objective from traitors
/:cl:

### DIFF
--- a/code/datums/antagonists/datum_traitor.dm
+++ b/code/datums/antagonists/datum_traitor.dm
@@ -157,9 +157,13 @@
 			break
 
 	if(martyr_compatibility && martyr_chance)
-		var/datum/objective/martyr/martyr_objective = new
-		martyr_objective.owner = owner
-		add_objective(martyr_objective)
+		//Nothing happens.
+		//This used to be the martyr objective to end your life,
+		//but since that doesn't give you a license to grief and
+		//some people might not even want to die,
+		//we're just leaving this empty.
+		//That's ltierally the best way to handle this.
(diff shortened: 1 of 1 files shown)


The-Driker:
Yes please.

Quantum-M:
I agree. People tend to presume that they can severely damage and murderbone the station.

Not many people are interested in killing themselves either.

Wilchenx:
image

Damn i wonder if there is any connection here 🤔

SamHPurp:
Conceptually: LGTM.

Didn't review the code though.

PR #1063, opened 2018-07-14
No More Green Slimes

AndrewMontagne:
Removes green slimes from the random mutation chance list. This is why we can't have nice things.

:cl: AndrewMontagne
tweak: Removed Green Slimes from the mutation chance table
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/slime/slime.dm
+++ b/code/modules/mob/living/simple_animal/slime/slime.dm
@@ -73,7 +73,7 @@
 
 	var/static/list/slime_colours = list("rainbow", "grey", "purple", "metal", "orange",
 	"blue", "dark blue", "dark purple", "yellow", "silver", "pink", "red",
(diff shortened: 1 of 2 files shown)


Battletrap360:
:(

kingoftheautism:
No fun allowed boys

Quantum-M:
I think you should make it just create water for now, as it will end up blocking off the black slimes, which can only be gotten from greens.

SamHPurp:
This is going to have negative knock on effects. I get what you're trying to fix, but this isn't the way we're going to be able to do it.

alexkar598:
It's not that bad tbh. The bad ones should be removed instead(plasma men and vox)

Quantum-M:
@alexkar598 Issue is that idiots go and meme with the mutation toxins, especially the Ethari one.

alexkar598:
Remove ethari Mut toxin then,weren't ethari supposed to be removed?

PR #1072, opened 2018-07-18
General animation fancification

### LINKED ISSUE #897: Lavaland Mysterious Boat Lava bug
[Note]: # (<B>Please enter the `Round ID` from the `Status` panel if you can!</B> If you believe the issue to be caused by a testmerge, please report it in its relative PR thread. State what the issue is from a "whats wrong" prospective. Issue reports should clearly allow maintainers to understand whats wrong and how to test/reproduce if that is not obvious. Avoid ambiguity. Start your issue report below both of these lines (or remove them))

[Admins]: # (If you are reporting a bug that occured AFTER you used varedit/admin buttons to alter an object out of normal operating conditions, please verify that you can re-create the bug without the varedit usage/admin buttons before reporting the issue.)
Round ID: 2243

Whats Wrong?: I was riding the "Mysterious Boat" loot from lavaland in the lava and when I brought my kinetic accelerator and my advanced plasma cutter to my hands, they caught fire despite me being in the boat.
How to replicate this issue?: Get a Mysterious Boat, get in it, move to a lava tile and pull out a kinetic accelerator or an advanced plasma cutter.

I assume it'll work with any item but I wanna avoid any ambiguity and give you exactly what/how it happened.

FlattestGuitar:
:cl: Flatty
tweak: The attack animations will be just a bit prettier now.
feature: There are pickup animations for items now! You can go around them by turning Walk mode on.
/:cl:

I made a really weird change to `code/game/objects/items.dm`, which, even though fairly logical, has the potential to break shit. I've tested it quite a bit and it seems to work without hiccups, but please read that code carefully and tell me whether it will break everything or not. K thx. That also fixes #897, because why wouldn't it, obviously.

ALSO, I think this might be broken by Purp's big-ass /tg/ port, so I'd like that merged before this PR. :candle:

Also ports some PRs from /tg/ that are related.

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -264,7 +264,7 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 	if(istype(loc, /obj/item/storage))
 		//If the item is in a storage item, take it out
 		var/obj/item/storage/S = loc
-		S.remove_from_storage(src, user.loc)
+		S.remove_from_storage(src, S.loc)
 
 	if(throwing)
 		throwing.finalize(FALSE)
@@ -286,7 +286,7 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 
 	if(istype(loc, /obj/item/storage))
 		var/obj/item/storage/S = loc
-		S.remove_from_storage(src, user.loc)
+		S.remove_from_storage(src, S.loc)
 
 	if(throwing)
 		throwing.finalize(FALSE)
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -698,3 +698,31 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 /obj/item/MouseExited()
 	deltimer(tip_timer)//delete any in-progress timer if the mouse is moved off the item before it finishes
 	closeToolTip(usr)
+
+/obj/item/proc/do_pickup_animation(turf/target)
+	set waitfor = FALSE
+	var/turf/T = get_turf(src)
+	var/image/I = image(icon = src, loc = loc, layer = layer + 0.1)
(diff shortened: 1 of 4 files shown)


code/game/objects/items.dm
```diff
@@ -269,13 +269,13 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
...
 	if(loc == user)
-		if(!user.dropItemToGround(src))
+		if(!user.temporarilyRemoveItemFromInventory(src))// changed this line from dropItemToGround() to this; not sure why it works, but it works ~Flatty
 			return
 
 	pickup(user)
 	add_fingerprint(user)
 	if(!user.put_in_active_hand(src))
-		dropped(user)
+		src.forceMove(user.loc)// this line had dropped() on it and I am also not sure why this works but it does ~Flatty
```
FlattestGuitar: `temporarilyRemoveItemFromInventory()` already handles `dropped()`, so this shouldn't be an issue.

DaveTheHeadcrab:
Important note: This makes stealthily picking up any item pretty difficult under pretty much any circumstance. Not necessarily a problem, but definitely something to be aware of. Might want to add a system for quietly grabbing items.

Ziiro:
Easy solution: Make pickpockets gloves not play this animation when worn. Gives extra value to an undervalued piece of equipment.

AdamElTablawy:
@Ziiro those gloves don't exist

FlattestGuitar:
While they could be added, I don't think it's a good idea. Anyone would be instantly able to meta that you have them on if they saw no animation.

Someone on /tg/ pitched the idea of walk mode enabling you to do this silently and I'm in love.

FlattestGuitar:
You can now pick the items up without an animation if you're in walk mode.

PR #1083, opened 2018-07-22
the clown can't honk if you disable his job slot

FlattestGuitar:
:cl: Flatty
tweak: You now need to emag an ID console to use it to open Clown and Mime job slots.
/:cl:

Not a problem for a traitor to do this. 

No longer a problem for the administration team.

Win-win. Instant merge please.

### DIFF
--- a/code/game/machinery/computer/card.dm
+++ b/code/game/machinery/computer/card.dm
@@ -40,6 +40,11 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
 		"Brig Physician",
 		"Internal Affairs Agent")
 
+	var/list/emag_blacklisted = list(
+		"Clown",
+		"Mime"
+		)
+
 	//The scaling factor of max total positions in relation to the total amount of people on board the station in %
 	var/max_relative_positions = 30 //30%: Seems reasonable, limit of 6 @ 20 players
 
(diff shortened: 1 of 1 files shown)


AffectedArc07:
speedmerge

Ziiro:
Make it instantly open two slots please. Having to sit at the console and open those slots one by one with a timer is a pretty easy way to get caught

Ralta:
@Ziiro The only way that'd be acceptable is if it also had some form of emagging 'animation' (e.g. sparking like a door/APC), otherwise there's no indication it's being emagged like most other things. Actually, do supply consoles do a thing when emagged? I can't remember... :thinking: 

One question on this - does it allow remove the access requirement for the console? If it doesn't, then that makes it pointless for anyone but the HoP anyway. Should the emag should both open up the console for anyone to use? This might end up being OP, since giving yourself all access would be very easy for all traitors at that point. Hm. Maybe @Ziiro 's idea (immediately opening two slots) is better?

FlattestGuitar:
This doesn't bring down ID requirements. That would be lethal. Literally anyone with an emag would automatically break into any of the places with an ID computer and break the station in half.

Same goes for bringing the time lock down, in my opinion. Way too easy to do it super quickly.

Ziiro:
Fair, actually. Disregard my ideas.

Iamgoofball:
This is stupid.

PR #525, opened 2018-02-02
Ports Paradise ID Card Sprites

AffectedArc07:
:cl: AffectedArc07
add: Ported the paradise ID card sprites
tweak: Jobs now spawn with appropriate department ID cards
/:cl:

**IGNORE THE FIRST COMMIT, I BROKE MY FORK OK?!**

All jobs now spawn with relevant ID cards, Some of the sprites are unused, as this is a direct rip from para, in terms of sprites. This has been tested and it works.

Before:
image

After:
image

PR #1084, opened 2018-07-22
Adds ID painting to the PDA painter

AffectedArc07:
The PDA painter is now the `color manipulator`, as it can now edit IDs.
This and the PDA list both have "friendly" names, instead of state names.

Sprite changes:
https://imgur.com/a/NJrcSKI

:cl: AffectedArc07
add: You can now skin IDs in the PDA painter
imageadd: The PDA painter sprite now has a glass slot to insert IDs to be painted
tweak: The "PDA painter" is now called "color manipulator"
tweak: You can now emag the PDA painter to paint any style of PDA or ID (apart from admin ones)
tweak: The PDA painter will now show "friendly" names for PDAs and IDs instead of icon state names
fix: You can now paint a PDA in any station PDA scheme
imageadd: The augment manipulator has its own sprite (Done by Quantum_M)
/:cl:

Why: We have needed a way to paint IDs since #525 and I feel that I could combine it into this. The code for this is also *slightly* less shitcode.

Huge thanks to flatty and dave for making this actually work.


code/game/machinery/PDApainter.dm
```diff
@@ -144,6 +207,20 @@
 	else
 		to_chat(usr, "<span class='notice'>The [src] is empty.</span>")
 
+/obj/machinery/pdapainter/verb/ejectid()
```
FlattestGuitar: These verbs are a crime against UX. Remove them and add an Alt+Click eject. Also put it in the description of the thing, in bold.

`\n<b>Alt-Click</b> to eject its contents!`

FlattestGuitar: Also make them one proc, not two.

code/game/machinery/PDApainter.dm
```diff
@@ -1,66 +1,106 @@
...
-		D.name = D.icon_state //PDAs don't have unique names, but using the sprite names works.
-
-		src.colorlist += D
-
 /obj/machinery/pdapainter/Destroy()
 	QDEL_NULL(storedpda)
+	QDEL_NULL(storedid)
 	return ..()
 
 /obj/machinery/pdapainter/on_deconstruction()
```
FlattestGuitar: Once you make that single eject proc, use it here so as not to write the same code multiple times.

code/game/machinery/PDApainter.dm
```diff
@@ -1,66 +1,106 @@
...
+	desc = "A machine able to color PDAs and IDs with ease. Insert an ID card or PDA and pick a color scheme."
 	icon = 'icons/obj/pda.dmi'
-	icon_state = "pdapainter"
+	icon_state = "coloriser"
 	density = TRUE
 	anchored = TRUE
 	var/obj/item/device/pda/storedpda = null
-	var/list/colorlist = list()
+	var/obj/item/card/id/storedid = null
+	var/pda_icons = list(
```
FlattestGuitar: While the ID list is fine, the PDA list can be automated fairly sensibly. 
Put this in the init proc.
```
for(var/path in subtypesof(obj/item/device/pda)
    var/obj/item/device/pda/P = new path
    if(!P.hidden)
        pda_icons[P.name] = P.icon_state
    qdel(P)
```

code/game/machinery/PDApainter.dm
FlattestGuitar:
```
/obj/machinery/pdapainter/AltClick(mob/user)
* range check *
* incapacitated check *
* eject PDA *
* eject ID *
* mucho bene *
```

code/game/machinery/PDApainter.dm
AffectedArc07: Currently it goes by jobs, and I would rather it be standard to job than have PDA tacked onto it, as that was the reason to change it in the first place

code/game/machinery/PDApainter.dm
AffectedArc07: There already are
`ejectid()`
`ejectpda()`

PR #1086, opened 2018-07-23
Changes the snappop shuttle's chasm room with a brig.

Quantum-M:
Changes the snappop shuttle's chasm room with a brig.
:cl: 
tweak: Snappop(tm) no longer has a unsafe, literal bottomless pit room for clowns to fall into. It has been replaced with a brig following SOSHA guidelines.
/:cl:

The shuttle has no brig, and this room only encourages people to thrown in other people (specifically antags), this room is replaced with a regular brig, along with a airlock connecting to the docking port.

snappopshuttle

### DIFF
--- a/_maps/shuttles/emergency_clown.dmm
+++ b/_maps/shuttles/emergency_clown.dmm
@@ -103,7 +103,8 @@
 /turf/closed/wall/mineral/titanium,
 /area/shuttle/escape)
 "au" = (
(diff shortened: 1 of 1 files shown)


Cactus2388:
but where will I get muh greentext now?

Ralta:
I loved the chasm, but this is just more sensible for us.

PR #1091, opened 2018-07-23
Newer Health HUDs

FlattestGuitar:
Mostly meant as an alternative to #957. 

I'm torn on this myself, because it takes away from the functionality of medHUDs without a solid justification, but it will make us more unique and (at least in my opinion) these look way better.

If anyone likes these, I can make the rest of the HUD match in style. The ECG bit was definitely the hardest part.

:cl: Flatty
tweak: NEW MEDHUD SPRITES
/:cl:


Xantholne:
Tbh I was considering closing #957 to open a new one with the EKG graph style. I'm just bad at remembering to do things.

Rumiluntti:
Looks annoying and unnecessary.

Cactus2388:
This should probably get the [Stale] tag on it. I don't think it's an improvement really and updating the current one we have seems unnecessary.

miniusAreas:
I feel like a combo of both the new medihud and the old one would feel nice.

PR #1096, opened 2018-07-23
Preserve supply console contraband/emag value

### LINKED ISSUE #740: Hacked Cargo console does nothing
Note: This is distinctly different from an emagged cargo console.

For those who do not know what I am talking about, you can deconstruct a cargo console until you have removed its board. You can then multitool the board (change its spectrum to Broad) and then reinstall it into the console. The newly rebuilt cargo console should have an expanded list of items which you can order.

I've tried to redo this process again (when the menu of items which can be ordered has not changed) and have found multitooling the same board to ask me if I want to change to Broad again. This suggests the board sorts of resets after being reinstalled (because on the initial point of changing it to broad, multitooling it again will ask me if I want to make it standard) I don't know the actual problem so this is only an assumption.

tjg1:
Reassembling a supply console no longer resets the emagged/contraband values back to 0. 

This accounts for the supply console being unique in that its circuit board can be taken out and hacked/emagged while out of the computer frame.

Fixes #740

:cl: rd
fix: Supply console circuit boards preserve their contraband/emagged value on reassembly
/:cl:

### DIFF
--- a/code/modules/cargo/console.dm
+++ b/code/modules/cargo/console.dm
@@ -21,9 +21,10 @@
 /obj/machinery/computer/cargo/Initialize(mapload, obj/item/circuitboard/computer/cargo/C)
 	. = ..()
 
-	// Propagate contraband/emag values from the circuit board used to construct the console.
-	contraband = C.contraband
-	emagged = C.emagged
+	if (C)
+		// Propagate contraband/emag values from the circuit board used to construct the console.
+		contraband = C.contraband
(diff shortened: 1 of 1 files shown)


Keluandrie:
I've concerns over this for both sides, antag and non. Personally, after ordering nulls, I reassemble to ensure my cover isn't blown. After this, even if the perpetrator has been caught or is done, nullcrates are just on the menu permanently on that console, short of destroying it.

PR #1097, opened 2018-07-24
Dave's Big Fat Smelly Ugly Storage Rebalance: MK III - Revengeance

Dave-TH:
:cl:
add: A second belt slot now exists. It works like the first one, except you can only have one full belt on either slot at a time.
del: Removes the suit storage and PDA slots.
tweak: Bags will now delay items being removed from them based on the kind of bag. You can avoid this by holding the bag in your hand when removing the item.
/:cl:

**OH MY GOD DO NOT MERGE THIS YET, ARE YOU CRAZY?! YOU'LL KILL SOMEONE WITH THIS SHIT!**

The first feature of this PR is the removal of the suit storage and PDA slots. The suit storage slot was inconsistant and spatially ambiguous and the PDA slot was weirdly specific for what it was on top of the spatial ambiguity. The two slots have been replaced with a second belt slot which functions exactly like the first one. An important feature of this is that you can only wear one full belt across both slots.

One major design note here is that large air tanks now do not have a suit storage slot to hide in. This is intentional. Yellow emergency airtanks last for 30+ minutes without adding additional air to the starting pressure or adjusting release settings, which is a perfectly reasonable amount of time to get your shit done and refill. The argument is that if you want an air tank that lasts all round, you should need to use your back slot.

Here's what the UI looks like now:
newui

The second thing this PR does is add a delay to removing items from backpacks and similar items when they are equipped to your back. The exact amount depends on the exact backpack. I'll put a table below. The design intent here is to differentiate between the kinds of backpacks and add some difficulty to pulling an entire armory out of your ass while on the move.

Another important design note here is the increased importance of belts. People working in their jobs field will have an advantage because they will have belts available that grant quick access to job-related materials. De-homogenizing jobs is good for encouraging interdepartmental co-operation.

Starting "survival" boxes are now smaller, and hold only 4 items.

To compliment this system, I've added a system where your backpack will deplete slots based on how much storage space it has left (This is mostly useful for satchels and other low-storage bag solutions). See below.

33905107-8a2672fa-df4b-11e7-8ce4-aca316ba92f4
33905110-8b60d84a-df4b-11e7-8a43-cd0d809aa788
33905111-8c24b0ee-df4b-11e7-9c5c-cfd7b294fc1e
33905112-8d25d306-df4b-11e7-976a-e9c8f8250835
33905113-8deb366e-df4b-11e7-8f1f-c374e5deba4a

Storage capacity and item retrieval delay by bag type:
(Note: "Slots" refers to the number of w_class 3 items that can be fit into the bag. 3 w_class 1 items could be fit into a single "slot". Item delay is in seconds.)

| Bag Type  | Slots | Item Delay |
| ------------- | ------------- | ------------- |
| Duffelbag  | 10  | 5  |
| Backpack  | 7  | 3  |
| Messenger Bag  | 5 | 1 |
| Satchel | 4  | 0 |

To Do:
 - [ ] Test literally everything??
 - [ ] Remap some air tank locations in accounting for the new internal economy
 - [ ] Fix hardsuit jetpacks
 - [ ] Maybe consolidate some of the repeated belt code into convenient procs?
 - [x] Verify all of the mirror belt icons look pretty
 - [x] Maybe get a testmerge going and get feedback on the removal times for items
 - [x] I think this breaks some kinds of cardboard boxes. Verify this.
 - [ ] ~~Convince Purp that merging this is a good idea.~~ Raise enough money for a bribe. Kickstarter soon.


Keluandrie:
The madman did it.

DaveTheHeadcrab:
I'll probably edit the OP to be more coherent in the morning. I just wanted to get this PR posted. Will resolve conflicts later.

Ziiro:
Big 👍 from me conceptually. I'll have to actually play it to judge properly, but I like the idea.

FlattestGuitar:
I think the satchel would use a small delay as well for the consistency of people not pulling tazers out of their ass.

Also, possibly a message for when they do start digging in the bags?

Also, PRs like these kill puppies.

Ralta:
Cool idea - any chance there could be a different time taken to pull items out based on their size? i.e. the larger the item, the longer it takes. Or perhaps certain item 'types' (weapons, tools and the like) should take longer by virtue of being more bulky.

In any case, I'd love to see this test-merged in the **Near Future™**.

oversane:
time to swap to duffelbags huehuehue

Ziiro:
`[user] is pulling something out of their [bag]..`
`[user] is [rifling/rummaging] through their [bag].. `

AffectedArc07:
This isnt about backpacks at all. This is just you trying to remove the PDA slot again

🗞

Ziiro:
And he is correct to do so

kingoftheautism:
RIP Suit Storage, now forced to put warden's combat shotgun on backpack slot instead of suit storage, so warden will basically get straight downgrade of combat shotgun instead of ability to put it in a suit storage for a price of getting less ammo capacity

Also now forced to carry neither extra capacity o2 tank as miner (even filled they cannot last really long) or forget about backpack slot and put o2 tank on it

Ralta:
A filled emergency O2 tank lasts AGES. Dave seems to think they last 30+ minutes which, in my experience as an engineer, is probably right.

One filled emergency (yellow) tank should be more than enough for a miner to do one or two mining trips without refilling, easily. Stick it in your pocket like Vox et al do - it's not so bad once you're used to it (or stick it in the belt slot, which I assume still works).

Also, I'm 100% okay with the Warden not having constant easy access to their shotgun. It's compact, which means it will fit in a bag, unlike most other readily-available shotguns. But this applies to almost every other weapon too - eguns and tasers, for example. But you'll still have your Sec belt for flashbangs, stun batons, flashes etc.

kingoftheautism:
@Ralta I think compact combat shotgun still doesn't fit in a backpack. Also it's not enough for me since i'm forgetting to turn off internals once they on, and walking with them turned on 80% of the time making myself safe from BZ or small amounts of plasma in air or superheated/supercooled air

Ralta:
@kingoftheautism I'd have to check, but I assume that if it was able to fit in the suit slot, it should fit in a bag. Either way, I'm personally not particularly concerned.

As for the oxygen, it's ultimately up to you to ensure you turn your oxygen off, or refill it regularly. I have been known to forget to turn off my oxygen, so the game punishes you for neglecting to do so (as it should, imo). Frankly, I think this is a positive thing; it should make you more aware of the fact that you are wasting oxygen by wandering around the station with your mask up 24/7.

FlattestGuitar:
Something that was brought up was that the Warden's compact shotty will be useless now. Make it fit in a backpack.

Keluandrie:
Or replace it with the cycler shotgun, it has the same space requirement as combat and is pretty unique. Also, botany has sacks for plants that are unaffected by this.

Cactus2388:
change bag of holdings to have the same delay of the dufflebag. You wouldn't be able to pick something out of an infinite bag faster than you would a simple dufflebag. Also nerfs scientists running around with absolutely everything in a bag

FlattestGuitar:
press f for oracle

Ziiro:
f

PR #1101, opened 2018-07-24
Allow the light replacer to recycle objects directly off the floor

### LINKED ISSUE #1039: Borg Light Replacer Cannot Use Glass
The cyborg light replacer cannot recycle glass, even though it's description states it can.

The functionality should be fixed or the description amended.

tjg1:
This PR allows engineer borgs (and players) to feed glass/lightbulbs into their light replacer directly off the floor.

This was already working fine for `/obj/item/shard`, so I copied the code from `/obj/item/shard/attackby` to the other types of objects that can be fed into a light replacer.

Closes #1039 

:cl: rd
fix: Light replacers can now be fed materials directly off the floor. 
/:cl:

### DIFF
--- a/code/game/objects/items/storage/boxes.dm
+++ b/code/game/objects/items/storage/boxes.dm
@@ -531,7 +531,7 @@
 	use_to_pickup = 1 // for picking up broken bulbs, not that most people will try
 
 /obj/item/storage/box/lights/attackby(obj/item/I, mob/user, params)
-	if(istype(I, /obj/item/device/lightreplacer))
+	if(issilicon(user) && istype(I, /obj/item/device/lightreplacer))
 		I.attackby(src, user)
 	else
 		..()
--- a/code/game/objects/items/storage/boxes.dm
+++ b/code/game/objects/items/storage/boxes.dm
@@ -530,6 +530,13 @@
 	max_combined_w_class = 21
 	use_to_pickup = 1 // for picking up broken bulbs, not that most people will try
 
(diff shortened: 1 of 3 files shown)


code/game/objects/items/storage/boxes.dm
```diff
@@ -530,6 +530,13 @@
 	max_combined_w_class = 21
 	use_to_pickup = 1 // for picking up broken bulbs, not that most people will try
 
+/obj/item/storage/box/lights/attackby(obj/item/I, mob/user, params)
```
FlattestGuitar: You're strictly defining this for the `box/lights` subtype, which doesn't look like it's different from a regular box, so that's going to be confusing. Bad design! Move this proc to a regular `/obj/item/storage/box`. I'd also add a check to make sure it's a cyborg making this thing happen, because a human trying to put a light replacer into the bulb box might happen.

`if(issilicon(I.loc))` <- cyborg/AI check, though an AI is unlikely to wield anything.

DaveTheHeadcrab:
Also pretty sure the borg module is intentionally not able to recycle. It gets more charges by going in the recharger.

tjg1:
@FlattestGuitar `/obj/item/storage/box/lights` has its own unique sprite and can only hold lightbulbs. It can't hold a light replacer.

@DaveTheHeadcrab The cyborg light replacer could already recycle broken glass `/obj/item/shard`s. This extends it to other object types that can be fed into a light replacer by organics.

PR #1105, opened 2018-07-24
Oracle-y station names

FlattestGuitar:
The .json files for the different maps we have will now serve as sources of station names. The Captain's charter has been removed.

We'll probably need to talk about which names we want to have for each of the stations. This is also a great opportunity to gut all the stations we dislike, because that's easier than coming up with a coherent name.

:cl: Flatty
del: The Captain's charter has been removed. No more meme names. ;c
add: The station names are now locked in place and fancier, I guess. Hopefully.
/:cl:


spoonitate:
From the title of this I thought the name would be randomized every round. The charter was only really good for meme names, anyway.

Quantum-M:
I think we should keep New Harbour as a name for one of the stations, maybe metastation as it's used often?

Ralta:
Could we not keep New Harbour as the default and instead have the Captain's Charter bring up a list of alternative names to choose from? That way, the Captain can still change the station name to something else, but can't do stupid / rule-breaking names.

Ralta:
Because I'd rather they had _some_ level of choice (the list can be as long as you'd like, too) than literally zero choice. It kinda ties into the whole idea of changing instead of removing.

Your call though my dude, just a suggestion.

AffectedArc07:
Add a config file where you can put station names in that will be randomly picked

kingoftheautism:
@AffectedArc07 Soviet Fat Camp XII

Quantum-M:
Suggestions:

Metastation - New Harbour
Omegastation - Frontier Outpost
Pubbystation - Stella Station
Deltastation - West Point
Boxstation - Oraculi Station

PR #1108, opened 2018-07-25
Kills skateboard and scooter recipes

FlattestGuitar:
:cl: Flatty
del: You can no longer craft scooters and skateboards in the crafting menu
/:cl:

Because the new meta is using these at all times, apparently. They make you unslippable and fast at the cost of 10 metal. Looking like a shitty meme included.

You can still find the vehicles in different places all over the world.


AffectedArc07:
So megafauna are unkillable now?

oversane:
**WE MINERS NEED OUR SCOOTERS!!**

Issue #1003, opened 2018-06-23
Wallet and mining ID


Tatsu500:
Issue reported from Round ID: 3034 (Oracle Station | Medium RP)

This one is simple, i was a free golem and i made a wallet, when i put my mining id in it, both of them dissapeared. I don't think its a visual bug, because i tried cliclking on the ground to see if i was invisible, but nope, they just vanished.

FlattestGuitar:
You can't click invisible objects, because they have no hitbox. You can alt+click on a tile to see a list of its contents (or just right click). Have you done one of the latter to make sure that the wallet was not there? Alternatively, if it was in your hand or the ID slot. We'd want to know what happens there, exactly.

Affurit:
The wallet wasn't dropped. The issue seems to be that don't have sprites for many of the new ID cards that we've sprited, so as a result of that, when an ID is placed into the ID slot, it goes invisible as the icon is changed into something that doesn't exist? This happens with station ID cards too. 

Later insight after some testing: It happens with all but 2 ID cards on the station.

https://imgur.com/a/fRWkqms

Here you go, flatty.

tjg1:
I'm going to make the wallet `icon_state` fallback to `wallet_id` to stop it from disappearing.

Still needs proper sprites.

AndrewMontagne:
Duplicate of #837

PR #1110, opened 2018-07-26
Wallet icon_state fallback

### LINKED ISSUE #837: Wallet-ID bug
Round ID: 1713

Pick up wallet
Put ID in wallet
Wallet automatically equips to the ID slot?
Wallet is invisible now
walletmissing
Take ID out of wallet
visiblewallet
When I was wearing the ID in the wallet I couldn't examine myself to see I had the wallet on.

tjg1:
Prevents wallets from disappearing if there is no sprite matching the icon_state of ID inserted.

Fixes #837 #1003 

We might want to open a separate issue for creating wallet sprites matching the new ID sprites.

:cl: rd
fix: Wallets no longer disappear when IDs are inserted into them
/:cl:

### DIFF
--- a/code/game/objects/items/storage/wallets.dm
+++ b/code/game/objects/items/storage/wallets.dm
@@ -59,7 +59,10 @@
 /obj/item/storage/wallet/update_icon()
 	icon_state = "wallet"
 	if(front_id)
-		icon_state = "wallet_[front_id.icon_state]"
+		if ("wallet_[front_id.icon_state]" in icon_states(src.icon))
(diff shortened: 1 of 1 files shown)


FlattestGuitar:
Not the way to do it. BYOND already does this if you put an icon in the file without any icon state, so without a name. 

Also, I really want to remove wallets because they're silly and broken.

Citinited:
Wouldn't a better way to do it just be to overlay the ID's sprite onto the wallet?

PR #1114, opened 2018-07-29
Removes baseball bat recipe

Quantum-M:
Removes the baseball recipe from the wooden stack.
:cl: 
del: Removes baseball bat recipe from wooden stack.
/:cl:

In my opinion, the bats are only used to greytide, and isn't usually used by antags (since they already have tools at their disposal, usually) and if there is a ton of bats around the station, usually expect someone beating up someone else for no good reason.

**Tested**

### DIFF
--- a/code/game/objects/items/stacks/sheets/sheet_types.dm
+++ b/code/game/objects/items/stacks/sheets/sheet_types.dm
@@ -175,7 +175,6 @@ GLOBAL_LIST_INIT(wood_recipes, list ( \
 	new/datum/stack_recipe("honey frame", /obj/item/honey_frame, 5, time = 10),\
 	new/datum/stack_recipe("ore box", /obj/structure/ore_box, 4, time = 50, one_per_turf = TRUE, on_floor = TRUE),\
 	new/datum/stack_recipe("wooden crate", /obj/structure/closet/crate/wooden, 6, time = 50, one_per_turf = TRUE, on_floor = TRUE),\
-	new/datum/stack_recipe("baseball bat", /obj/item/melee/baseball_bat, 5, time = 15),\
 	))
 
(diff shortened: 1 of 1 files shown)


Ledwin1233:
That doesn't really make much sense as a reason to remove a way to get the item entirely though.

spoonitate:
Yeah, the baseball bat is almost entirely a greytide/self antag weapon. It's super annoying when someone hits someone else with a baseball bat in a crowded space, knocking several people over, which then leads to a room full of angry people looking to swat each other across the room.

Alternatives to removing the recipe:
- **Nerf the damage** - lowering the damage of the baseball bat seems like the obvious solution, but this could lead to people paradoxically using them more often because "it doesn't do that much damage, anyway".
- **Increase the production cost** - making the baseball bat more expensive to produce would do little to dissuade people from making them; while upping the resource cost would make it less convenient for someone to get the materials for it, the average greytider could just deconstruct the entire library for free.
- **Make them a traitor item** - jumping the shark and making the baseball bat a traitor item could lead to _fun things_ like justifiably exaggerating the knockback and damage, but that would eventually lead to people with baseball bats becoming valid salad.
- **Make a baseball bat crate with nothing but baseball bats** - Funny to think about, but cargo already has a problem with people busting crates of security gear open for personal use. 

All in all I'm glad the baseball bat is becoming uncraftable.

FlattestGuitar:
I feel like making it a traitor item doesn't make a whole lot of sense. What's the use in the baseball bat anyway? Does anyone use this outside of just greytide?

Quantum-M:
Unlike the other tools greytide can make, if you have a decent amount of wood, especially if you are botany, you can just spam bat create around the station. Why do people do this, is beyond me.

PR #1122, opened 2018-08-03
Infiltration mode!

steamp0rt:
:cl: steamp0rt
add: Infiltration gamemode
add: Basically some syndie agents have to activate _maximum stealth_ and sneak into the station to complete some objectives. They don't get access to certain weapons capable of causing mass harm/destruction and easily undoing any stealth.
add: They get 1 or 2 "unique" objectives, and 3-4 normal traitor objectives
add: The current unique objects are: steal X amount of power, kidnap a head of staff or security member, or hijack the AI
tweak: The Syndicate now realizes people may actually want to be stealthy with agent IDs, so they added the ability to disable the AI tracking jammer built into them.
code: Stuff for datumizing all the antags, which was needed to get Infiltrators to work
/:cl:

capture


steamp0rt:
I will note, this HAS been highly tested.

AndrewMontagne:
eyes2

AdamElTablawy:
traitor except the station instantly calls the shuttle when they see you because you're easily identified as infiltrators

good

steamp0rt:
@AdamElTablawy how? I don't even think even the AI actually checks the crew manifest, much less the crew

steamp0rt:
@AdamElTablawy ?

AdamElTablawy:
No, AI does check the crew manifest, though the simple fact that they won't show up as a listed job  in the radio channel will be enough of a giveaway for anyone with half a brain.

What's more is that it will be exceedingly obvious to notice them because they have unique gear and the moment one gets caught then it's a dead giveaway that they're an infiltrator, and it's an inevitability that at least *one* will get downed by security.

Also, if they're working in teams, then people will notice 2+ traitors together and shout 'INFILTRATOR.' if they aren't working in teams then again you have the same problem of one gets caught and the mode's instantly revealed.

this wouldn't be a problem, except that the moment the mode is revealed, the round is over, because everyone knows it's infiltrators now, not just traitors, because unless you are fantastically stupid it would be impossible to mistake them for just ordinary traitors. and because it's infiltrators they'll just call the shuttle and leave.

again, there's no real problem with the gamemode itself (aside from being fairly boring, but that's common for traitor in general, especially since the entire point of this gamemode is to NOT reveal yourself and thus if the infiltrators are great nothing will really be noticable), but it boils down to, again, traitor, except when one traitor gets caught, the shuttle gets called because it's infiltrator

code/datums/antagonists/infiltrator/items/hardsuit.dm
```diff
@@ -0,0 +1,36 @@
+/obj/item/clothing/head/helmet/space/hardsuit/infiltration
+	name = "security hardsuit helmet"
```
FlattestGuitar: Not really a security hardsuit helmet, though, is it

steamp0rt: The name will be changed as soon as the disguise changes, tbh. But i'll change it to default

spoonitate:
Agent IDs can change the visible job, yes, but the AI will be unable to track them even if they have a plausible name and job. A competent AI will notice this and immediately know that there are infiltrators aboard the ship. I, for example, will cross-reference people's names on the manifest and check for duplicate/missing biometric information in the medical records. 

While I think this is a neat concept in theory, in practice it would be incredibly hard to pull off without everyone deciding to collectively ease up on the validhunting to have a fun game of cat and mouse. Considering that the gamemode is all about stealth I feel like there should be on-base perks that you can purchase for TC that will make infiltration just that much easier, while having enough variety to trip up people attempting to meta the roundtype. Something like crew manifest spoofing, or having an advanced camera console on the ship for one infiltrator to run support surveillance, or maybe even faking centcomm announcements so people meta the wrong roundtype.

steamp0rt:
@spoonitate That's actually a good idea.

steamp0rt:
Also, there is already an advanced camera console on the ship.

Rumiluntti:
Vox raiders was a bloody mess, why do people insist on redoing the old "Like nukies but shit" thing

AndrewMontagne:
Playtested. Minor issues discovered but working overall. 👍

steamp0rt:
Turns out, the bug where rounds wouldn't end was due to the `check_finished` proc not automatically returning true if the shuttle has escaped. Whoops.

steamp0rt:
Should be merge ready

code/game/objects/items/implants/implant_misc.dm
```diff
@@ -83,3 +83,62 @@
...
+
+/obj/item/implant/radio/mining
+	radio_key = /obj/item/device/encryptionkey/headset_cargo
+
+/obj/item/implant/radio/syndicate
+	desc = "Are you there God? It's me, Syndicate Comms Agent."
+	radio_key = /obj/item/device/encryptionkey/syndicate
+	subspace_transmission = TRUE
+
+/obj/item/implant/radio/slime
```
FlattestGuitar: Probably pasted here by mistake.

FlattestGuitar: Actually, nevermind. Might as well leave it as an admin thing until this is used. Ignore that.

code/modules/uplink/uplink_item.dm
```diff
@@ -1335,6 +1351,63 @@ GLOBAL_LIST_EMPTY(uplink_items) // Global list so we only initialize this once.
...
+
+/datum/uplink_item/services/manifest_spoof
+	name = "Crew Manifest Spoof"
+	desc = "A button capable of adding a single person to the crew manifest."
+	item = /obj/item/device/service/manifest
+	cost = 15 //Maybe this is too cheap??
+
+/datum/uplink_item/services/fake_ion
+	name = "Fake Ion Storm"
+	desc = "Fakes an ion storm announcment. A good distraction, especially if the AI is weird anyways."
```
FlattestGuitar: `anyway`. The `s` is a grammatical error.

SamHPurp:
@AndrewMontagne we'll need to workout a weighting etc and stuff it into the config.

PR #1126, opened 2018-08-03
Lockboxes are harder to break

FlattestGuitar:
They're there for a reason. 20 shots from a mining tool shouldn't be enough to circumvent them.

:cl: Flatty
tweak: Lockboxes are way harder to break.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/lockbox.dm
+++ b/code/game/objects/items/storage/lockbox.dm
@@ -9,6 +9,8 @@
 	max_w_class = WEIGHT_CLASS_NORMAL
 	max_combined_w_class = 14 //The sum of the w_classes of all the items in this storage item.
 	storage_slots = 4
+	max_integrity = 1500
+	obj_integrity = 1500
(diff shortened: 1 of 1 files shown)


Ziiro:
To be fair, if they have mining implements, they have access to the mining shuttle. And if they have access to the mining shuttle they can just use lava.

spoonitate:
At least they can't just break it open in the middle of R&D.

FlattestGuitar:
They can print that mining laser that shoots the purple beam and do all that in the middle of RnD.

Quantum-M:
Stuff is secured for a reason. If you aren't a traitor or a security staff, why would you want to get a weapon in normal circumstances?

Ziiro:
Because there is no security 30 minutes into the shift, or the only security officer is dead

steamp0rt:
Or maybe there's a blob and the HoS/warden and too busy to bother

PR #1127, opened 2018-08-03
Removes socks.

FlattestGuitar:
:cl: Flatty
del: Socks have been removed from the game. I hate them.
/:cl:

Because:
* They're ugly
* They're bloat
* They don't work properly if you lose a leg and cause confusion because you can't see the leg is gone.
* Quite a few people don't use them altogether for those reasons.
* They're so minor that it's easier to remove them than bother.
* Socks killed my grandma.

There's no easy way to remove socks from our savefiles, so they'll stay there forever, but shouldn't cause much trouble.


steamp0rt:
#purgethesocks

Cactus2388:
but how will I keep my feet warm

AffectedArc07:
Flattest, you should be adding more socks, since you want barbers to have 3 pairs of shoes

Birdtalon:
>Socks killed my grandma.

Why did you even bother saying more? Remove them immediately.

Xantholne:
How can I have kinky ERP now though??????

spoonitate:
gotta remove these socks to let my GamerFeet breathe

PR #1129, opened 2018-08-03
Medbot faces!

FlattestGuitar:
Took me way longer than it should have. The colors aren't in place yet and I have to test all this, so DNM.

They have special expressions for a handful of different events, but I won't ruin that for you. The gif is purely idle emotions, which they do on..being idle.

:cl: Flatty
add: Your medbots might wink at you. Watch out.
/:cl:


FlattestGuitar:
You'll note the icons don't actually play the animations properly and cut them randomly. I consider this a feature, because they seem even more lifelike. I'm tearing up as I say this.

Quantum-M:
If a medibot winks at me, would I be able to take it to the IAA for harassment?

In seriousness, is it possible to give it a talking icon when it says stuff?

Cactus2388:
The yellow is really ugly

AdamElTablawy:
haha wow! this is hideous

FlattestGuitar:
That hurt my feelings. :c

AdamElTablawy:
these hurt my eyes

spoonitate:
While Adam was being pretty rude about it I do agree that the color is a bit offputting. Maybe turn it a shade of green like what I showed you a while back?

Rumiluntti:
Ebin smileycentral medibots.
Please no

Cactus2388:
Medibots are autonomous and don't need emotions. The piss yellow is off putting and doesn't match with anything else we have either.

miniusAreas:
this is great, yes.

FlattestGuitar:
I give up. Nobody likes my art. I'll commit a genocide instead. ;c

PR #1136, opened 2018-08-04
Doubles vacuum damage

ghost:
Ports https://github.com/tgstation/tgstation/pull/32566

🆑 steamport, originally by nicn
balance: Damage from low pressure has been doubled.
/🆑


spoonitate:
So the engineering department actually hauls ass to plug breaches I suppose?

steamp0rt:
@spoonitate I've actually spacewalked from the AIsat exterior to escape, on Meta, with just salbumatol pills and brute patches

kingoftheautism:
@steamp0rt Why not just use firesuit?

steamp0rt:
@kingoftheautism Because it was a wormhole teleport?

Ralta:
I'm okay with this. It makes you rethink space-walking in a big way, which shouldn't be nearly as 'easy' as it currently is.

PR #1140, opened 2018-08-06
bone fixing juice

FlattestGuitar:
:cl: Flatty
add: Syndicate agents can now purchase the C4L-Z1UM Agent for 3 TC. It's an autoinjector that will allow them to fix all broken bones and will mend all organs, however it will also put them to sleep for some time.
/:cl:

I got the sound from here: http://www.trekcore.com/audio/ its name is `Hypospray 2`. I'm not sure if we can actually use that, so I wrote to the people over there and maybe they'll respond one day. If not, we can steal a different one from somewhere else.


code/modules/reagents/chemistry/reagents/medicine_reagents.dm
```diff
@@ -1094,3 +1094,27 @@
 		return
 	var/mob/living/carbon/human/H = M
 	H.shock_stage = max(H.shock_stage - 1.5, 0)
+
+/datum/reagent/medicine/bonefixingjuice 
+	name = "C4L-Z1UM Agent"
```
Citinited: doot doot

code/modules/uplink/uplink_item.dm
```diff
@@ -1121,6 +1121,12 @@ GLOBAL_LIST_EMPTY(uplink_items) // Global list so we only initialize this once.
 	item = /obj/item/stack/telecrystal/five
 	cost = 5
 
+/datum/uplink_item/device_tools/telecrystal/five
```
Citinited: This datum type seems to be incorrect

Citinited:
looks like you're boned

PR #1142, opened 2018-08-07
Adds SOP books to metastation

Quantum-M:
Adds SOP books to metastation.

:cl: optional name here
add: Adds SOP books to metastation.
/:cl:

Some things were shifted around, largely useless clutter and folders to add the books.

### DIFF
--- a/_maps/map_files/MetaStation/MetaStation.dmm
+++ b/_maps/map_files/MetaStation/MetaStation.dmm
@@ -2897,11 +2897,11 @@
 	pixel_x = -3;
 	pixel_y = 3
 	},
-/obj/item/cigbutt/cigarbutt{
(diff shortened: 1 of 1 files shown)


_maps/map_files/MetaStation/MetaStation.dmm
```diff
@@ -2897,11 +2897,11 @@
 	pixel_x = -3;
 	pixel_y = 3
 	},
-/obj/item/cigbutt/cigarbutt{
```
SamHPurp: rip cigar butt and everything else that got removed instead of moved around

Quantum-M: I believe I only removed that cigar butt, two blue folders, and a blank paper.

PR #1149, opened 2018-08-09
Monkeys no longer remove your hair when worn

Ziiro:
:cl:
fix: Monkeys no longer remove your hair when worn
/:cl:


SamHPurp:
Un-realistic, I'd suspect an angry monkey would entirely remove all of your hair! 🤣

SamHPurp (approved):

PR #1152, opened 2018-08-17
Get rid of portable chem dispensers, and make the normal chem dispensers constructable and upgradable

steamp0rt:
:cl: steamp0rt
refactor: Chem dispensers are now actual constructable machines
del: Portable chem dispensers. There was really no point of them, besides them being constructable, but now that normal chem dispensers can do that, there's no point of that anymore.
/:cl:


Quantum-M:
Strongly disagree on the macro, chemistry is supposed to be tedious, not a quick process and it just makes it slightly more easier.

On the dispenser itself, the chemical and the energy the ones that start on the map are affected? Since a unupgraded chemical dispenser (the portable ones we have at the moment) are really weak on both chem numbers and energy numbers.

FlattestGuitar:
I will not stand for copypasting ANYTHING into the game as a part of normal gameplay. No.

I'd be fine with recording recipes so you'd at least have to make something once during the round, but this is broken garbage that will only encourage people to exchange some minmax guides on the Discord. No.

steamp0rt:
@Quantum-M Yes, ALL chem dispensers are now upgradable

@FlattestGuitar You can't make precise (read: good) macros without upgrades anyways

Cactus2388:
Chemistry is already stupid easy and overpowered. Macros aren't needed.

steamp0rt:
If this gets firmly denied, I'd still be willing to PR the upgradable chem dispenser alone.

steamp0rt:
OK, I removed macros, and just left the upgradable chem dispensers

steamp0rt:
Ready

PR #1161, opened 2018-08-22
Gives the Mime the same clumsy mutation as the clown

Quantum-M:
:cl: 
tweak: Mime now starts with the same mutations as the Clown.
/:cl:

My reasoning is this: The mime should at least be humorous/entertaining to the crew. Also, to discourage the ever so present mindset of almost all Mimes we get to greytide and be shitters.

The antag stuff was adjusted to taken in account the Mimes getting the mutations, as in its removed as they do currently with the Clown, but I wasn't able to find the part for traitor for the mutation removal.

### DIFF
--- a/code/modules/jobs/job_types/civilian.dm
+++ b/code/modules/jobs/job_types/civilian.dm
@@ -114,6 +114,8 @@ Mime
 		H.mind.AddSpell(new /obj/effect/proc_holder/spell/targeted/mime/speak(null))
 		H.mind.miming = 1
 
+	H.dna.add_mutation(CLOWNMUT)
+	H.dna.add_mutation(WACKY)
 /*
 Curator
 */
--- a/code/modules/jobs/job_types/civilian.dm
+++ b/code/modules/jobs/job_types/civilian.dm
@@ -115,7 +115,6 @@ Mime
 		H.mind.miming = 1
 
 	H.dna.add_mutation(CLOWNMUT)
-	H.dna.add_mutation(WACKY)
 /*
 Curator
 */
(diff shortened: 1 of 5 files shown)


code/modules/jobs/job_types/civilian.dm
```diff
@@ -114,6 +114,8 @@ Mime
 		H.mind.AddSpell(new /obj/effect/proc_holder/spell/targeted/mime/speak(null))
 		H.mind.miming = 1
 
+	H.dna.add_mutation(CLOWNMUT)
+	H.dna.add_mutation(WACKY)
```
FlattestGuitar: `WACKY` adds the comic sans voice, which I think isn't quite what you intended here.

Cactus2388:
I don't think this is the best way to remove mime-tide. The clown has the clumsy trait because clowns are famous for slapstick comedy. A mime being clumsy doesn't really make sense.

Quantum-M:
Mimes on some fictions are sometimes shown as clumsy (although said Mimes usually end up in the face of random, comical violence) and the mutation shouldn't really appear in a big manner unless they handle cuffs, guns or such, @Cactus2388

Quantum-M:
Comic sans mutation removed from the Mime.

Ralta:
We should give mimes their own actual gimmick. Make it so they can hardly use *anything* in-hand, but get more stupid abilities to make them more 'mime-like'.

```
e.g. 
Pulling an item with an invisible rope to drag it (*slowly* from a distance - 2 or 3 tiles away perhaps)
Mime being drunk, which *literally makes them drunk*
Mime various items, such as throwing a banana cream pie, slipping on a peel, spraying a seltzer bottle etc
If they have an item in their possession, they have to mime using it for it to do anything (a good mime doesn't need props)
```

Something. Anything that makes them unique, and not just a silent clumsy clown/greytider. At least the clown can be funny (and promptly beaten). Mimes are worthless without something that makes them interesting, in my opinion.

PR #1172, opened 2018-08-26
Shuttle Navigation Computer QOL changes.

AndrewMontagne:
This adds a "Travel" action while in the Shuttle Navigation Computer, provided it has the `possible_destinations` var populated. It also adds some code to automatically populate the shuttle vars on the console. The practical upshot of this is that the shuttle docker can be used to control a shuttle entirely on its own.

Finally, it now sets the client's scaling mode to stretch to fit whilst using the console, as in the case of the infiltrators shuttle, the view range increases; it sets the scaling mode back to the previous value once they exit the console.

:cl: AndrewMontagne
tweak: Quality of life improvements to the Shuttle Navigation Computer
/:cl:


kingoftheautism:
Can we get new whiteships (at least new meta whiteship) from /tg/ too? It will require some docking remapping however

